PR #159, opened 2021-03-03
Bugfix/NUI-5864 overlay does not close with dialog window in date time picker component

Vita-Karas:


packages/bits/src/lib/date-picker/date-picker.component.ts
```diff
@@ -302,6 +302,10 @@ export class DatePickerComponent implements OnChanges, OnInit, ControlValueAcces
             this.calendarChanged.unsubscribe();
         }
 
+        if (this.overlay.showing) {
+            this.overlay.hide();
+        }
```
Blakesalot, line 307: it would be good to add a unit test or some form of automation test for this

packages/bits/src/lib/time-picker/time-picker.component.ts
```diff
@@ -290,6 +290,10 @@ export class TimePickerComponent implements OnInit, OnDestroy, OnChanges, AfterV
     }
 
     ngOnDestroy() {
+        if (this.overlay.showing) {
+            this.overlay.hide();
+        }
```
Blakesalot, line 295: another instance where it would be good to add a unit test or some form of automation test

Blakesalot (changes requested):

packages/bits/src/lib/time-frame-picker/time-frame-picker.component.ts
```diff
@@ -38,6 +39,8 @@ export class TimeFramePickerComponent implements OnChanges, OnInit {
 
     public distanceToEndDate: number; // to keep distance between start and end-date
 
+    public onDestroy$ = new Subject<void>();
```
Blakesalot, line 42: Is this subject needed? I don't see where it's being used.

Vita-Karas: will removed, thanks

packages/bits/src/lib/date-picker/date-picker.component.ts
Vita-Karas: Will add, thanks

packages/bits/src/lib/time-picker/time-picker.component.ts
Vita-Karas: Will added, thanks

dmytrofirsov (dismissed):

packages/bits/demo/src/components/demo/dialog/dialog-inside-overlay/dialog-inside-overlay.example.component.html
```diff
@@ -12,11 +12,12 @@
     <nui-dialog-header title="Dialog Title" i18n-title (closed)="close()"></nui-dialog-header>
     <div class="dialog-body">
         <p>
-            <nui-select-v2 placeholder="Select an item" i18n-placeholder>
+            <nui-select-v2 class="mb-5" placeholder="Select an item" i18n-placeholder>
                 <nui-select-v2-option *ngFor="let option of options2" [value]="option">
                     {{ option }}
                 </nui-select-v2-option>
             </nui-select-v2>
+            <nui-date-time-picker class="date-time-picker" (modelChanged)="onModelChanged($event)" displayMode="inline" [model]="dt" [appendToBody]="true"></nui-date-time-picker>
```
Blakesalot, line 20: We should avoid adding things to our examples that are just for testing purposes. If we need a component for testing, do you mind creating a separate one? That way we can avoid cluttering up the example.

packages/bits/spec/components/dialog/dialog.atom.ts
```diff
@@ -41,4 +42,10 @@ export class DialogAtom extends Atom {
     public scrollToEnd = async (): Promise<void> => browser.executeScript<void>(`arguments[0].scrollIntoView({block: "end"})`, this.getDialog());
 
     public getDialog(): ElementFinder { return super.getElement().element(by.className("modal-dialog")); }
+
+    public getDatetimePicker(): DateTimepickerAtom {
```
Blakesalot, line 46: We should avoid adding this method to the dialog atom since a datetime picker isn't a basic/normal part of a dialog.

packages/bits/spec/components/dialog/dialog.atom.ts
```diff
@@ -41,4 +42,10 @@ export class DialogAtom extends Atom {
...
+        return Atom.findIn(DateTimepickerAtom, this.getElement().element(by.className("date-time-picker")));
+    }
+
+    public getBody(): ElementFinder { return super.getElement().element(by.className("body")); }
```
Blakesalot, line 50: I don't see where this "body" class is being specified in a template. Is this method needed? If so, I don't think the method should be included in the dialog atom since the body template is controlled by the consumer, and plus they can put whatever classes they want on it.

packages/bits/src/lib/date-picker/date-picker.component.less
```diff
@@ -78,10 +78,10 @@
         min-height: @nui-datepicker-popup-min-height;
         min-width: @nui-datepicker-popup-min-width;
         padding: (@nui-space-sm - @nui-space-xxs) @nui-space-sm;
-        .setCssVariable(background-color, nui-color-bg-light);
 
         table {
             outline: none;
+            background-color: transparent;
```
Blakesalot, line 84: Just curious, what is this change for?

Blakesalot (changes requested):

packages/bits/demo/src/components/demo/dialog/dialog-inside-overlay/dialog-inside-overlay.example.component.html
Vita-Karas: Yeah, I agree. Created a separate one for testing purposes, thanks

packages/bits/spec/components/dialog/dialog.atom.ts
Vita-Karas: missed that, thanks

packages/bits/src/lib/date-picker/date-picker.component.less
Vita-Karas: It is a fix for failed date-time-picker visual tests, which were found by Ivan.

packages/bits/demo/src/components/demo/dialog/dialog-docs/dialog-docs.example.component.html
```diff
@@ -131,6 +131,10 @@ <h2>Dialog Inside Overlay</h2>
     <nui-dialog-inside-overlay-example></nui-dialog-inside-overlay-example>
 </nui-example-wrapper>
 
+<nui-example-wrapper filenamePrefix="dialog-inside-overlaywith-date-time-picker" exampleTitle="Dialog Inside Overlay With DateTimePicker Component">
```
Blakesalot, line 134: I don't think we need an example in the docs for this since it's only a bugfix. We should make sure anything that goes in the docs is added there for the purpose of increasing the understanding of our consumers.

For the e2e tests, the best thing to do in this case will be to create a separate test component that instantiates all the same components that are instantiated in this existing DialogDocsComponent. It can be called DialogTestComponent. The "dialog-test" route found [here](https://github.com/solarwinds/nova/blob/f1291391005ea3e52a332abd99ce70729af1ef36/packages/bits/demo/src/components/demo/dialog/dialog.module.ts#L57) can point to this new DialogTestComponent instead of the docs one.

Blakesalot (changes requested):

packages/bits/demo/src/components/demo/dialog/dialog-docs/dialog-docs.example.component.html
Vita-Karas: Created a separate test component, thanks

Blakesalot (approved):

PR #164, opened 2021-03-10
Feature/NUI 5749 migrate nova bits to eslint

dmytrofirsov:


packages/bits/package.json
```diff
@@ -152,6 +159,11 @@
...
     "codelyzer": "^6.0.1",
     "core-js": "^2.6.9",
+    "eslint": "^7.6.0",
+    "eslint-import-resolver-typescript": "^2.4.0",
+    "eslint-plugin-import": "2.22.1",
+    "eslint-plugin-jsdoc": "30.7.6",
+    "eslint-plugin-prefer-arrow": "1.2.2",
     "graphql": "^15.4.0",
     "graphql-tag": "^2.11.0",
     "highlight.js": "^10.6.0",
```
Blakesalot, line 169: should tslint be removed at line 189?

packages/bits/src/lib/popup/popup-toggle.directive.ts
```diff
@@ -12,9 +12,9 @@ import {
     selector: "[nuiPopupToggle]",
 })
 export class PopupToggleDirective {
-    /** sets disable state and which prevents emitting toggle**/
+    /** sets disable state and which prevents emitting toggle* */
     @Input() isDisabled: boolean;
-    /** sets disable state and which prevents emitting toggle (for using it on textbox component)**/
+    /** sets disable state and which prevents emitting toggle (for using it on textbox component)* */
```
Blakesalot, line 17: extra asterisk

```suggestion
    /** sets disable state and which prevents emitting toggle (for using it on textbox component) */
```

packages/bits/src/lib/popup/popup-toggle.directive.ts
```diff
@@ -12,9 +12,9 @@ import {
     selector: "[nuiPopupToggle]",
 })
 export class PopupToggleDirective {
-    /** sets disable state and which prevents emitting toggle**/
+    /** sets disable state and which prevents emitting toggle* */
```
Blakesalot, line 15: extra asterisk

```suggestion
    /** sets disable state and which prevents emitting toggle */
```

packages/bits/src/lib/popup/popup-container.component.ts
```diff
@@ -26,7 +26,7 @@ import { PositionService } from "../../services/position.service";
     encapsulation: ViewEncapsulation.None,
 })
 
-/* tslint:disable:use-host-property-decorator */
+/* eslint-disable */
```
Blakesalot, line 29: is there a rule name that can be specified here?

packages/bits/src/lib/popup/popup.component.ts
```diff
@@ -45,7 +45,7 @@ import { PopupToggleDirective } from "./popup-toggle.directive";
  * 1st layer of content to define "popup area". DO NOT USE "opened" WITH "nuiPopupToggle".
  * IT IS NEEDED TO CHOOSE ONE.
  */
-/* tslint:disable:use-host-property-decorator */
+/* eslint-disable  */
```
Blakesalot, line 48: is there a rule name that could be specified here?

packages/bits/src/lib/popup/popup.component.ts
```diff
@@ -70,7 +70,7 @@ import { PopupToggleDirective } from "./popup-toggle.directive";
     styleUrls: ["./popup.component.less"],
     encapsulation: ViewEncapsulation.None,
 })
-/* tslint:disable:use-host-property-decorator */
+/* eslint-disable */
```
Blakesalot, line 73: is there a rule name that could be specified here?

packages/bits/demo/.eslintrc.json
```diff
@@ -0,0 +1,17 @@
+{
+    "extends": "../.eslintrc.json",
```
Blakesalot, line 2: it seems like some properties in this file can be omitted since they have the same values as the one its extending

packages/bits/demo/src/.eslintrc.json
```diff
@@ -0,0 +1,31 @@
+{
```
Blakesalot, line 1: is there a reason why the demo app needs two .eslintrc files? I wonder if they could be combined into a single file.

if two files are indeed needed, maybe some of the properties in this file could be omitted if they have the same value as the one being extended.

Blakesalot (changes requested):

packages/bits/package.json
dmytrofirsov: right, should have removed that

packages/bits/src/lib/popup/popup-container.component.ts
dmytrofirsov: good catch, the **@angular-eslint/no-host-metadata-property** rule it is. Added

packages/bits/demo/.eslintrc.json
dmytrofirsov: I think we can omit **"createDefaultProgram": true**, but the rest should be there, the "overrides" key requires the definition of "files" types

packages/bits/demo/src/.eslintrc.json
dmytrofirsov: .eslintrc.json files are generated for each angular project respectively. Since both "demo" and "demo-e2e" projects use the same "docs" root directory, but different source directories, they're generated in different places. They're only different with a specific tsconfig file, the rest of the properties are mandatory.

I tried to play around with the project source folders and config files and could only make it work the default way schematics generate those files. I need more time to investigate this, if it's a priority I'm fine with having a task to investigate.

packages/bits/123.eslintrc.js
```diff
@@ -0,0 +1,241 @@
+/*
```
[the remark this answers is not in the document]
dmytrofirsov: this one is indeed autogenerated, just forgot to remove it

ts-lint-rules/tslint.angular.json
```diff
@@ -1,11 +1,11 @@
 {
```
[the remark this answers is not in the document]
dmytrofirsov: Because we have a job **"dist-copy:tslint"** and we ship tslint.json with our package, which is actually now extends from a non-existing file. I think we should fix that first if we still want to deliver tslint rules to our consumers, someone can still be using those

packages/bits/src/lib/popup/popup.component.ts
dmytrofirsov: added

packages/bits/demo/src/.eslintrc.json
Blakesalot: Okay, the current setup is alright. I wouldn't consider figuring out a one-file solution to be a priority.

ts-lint-rules/tslint.angular.json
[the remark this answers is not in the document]
Blakesalot: @dmytrofirsov, is there a ticket for figuring this out?

Blakesalot (approved):

Vita-Karas (approved):

PR #167, opened 2021-03-10
NUI-5762/Error on table widget row click

baggiosw:
Created a test and fix for ticket

### DIFF
--- a/packages/dashboards/src/lib/components/providers/interaction/url-interaction-handler.spec.ts
+++ b/packages/dashboards/src/lib/components/providers/interaction/url-interaction-handler.spec.ts
@@ -15,7 +15,6 @@ describe("UrlInteractionHandler", () => {
         let handler: UrlInteractionHandler;
         let eventBus: EventBus<IEvent>;
         let window: any;
-        let loggerWarnSpy: Spy;
 
         beforeEach(() => {
             window = {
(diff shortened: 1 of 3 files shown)


packages/dashboards/src/lib/components/providers/interaction/url-interaction-handler.spec.ts
```diff
@@ -74,6 +75,14 @@ describe("UrlInteractionHandler", () => {
             expect(window.location.href).toEqual("default");
         });
 
+        it("logger should warn if properties are not loaded", () => {
+            // @ts-ignore
+            handler.updateConfiguration(undefined);
+            eventBus.getStream(INTERACTION).next({ payload });
+            expect(handleInteractionSpy).toHaveBeenCalled();
+            expect(loggerWarnSpy).toHaveBeenCalled();
+        });
```
Blakesalot, line 84: I think a more direct regression test for the bug would be something like this:
```
expect(() => eventBus.getStream(INTERACTION).next({ payload })).not.toThrow();
```

Similar to the test [here](https://github.com/solarwinds/nova/blob/f1291391005ea3e52a332abd99ce70729af1ef36/packages/charts/src/core/grid/xy-grid.spec.ts#L578).

Blakesalot (changes requested):

Blakesalot (approved):

PR #168, opened 2021-03-11
NUI-5695 dynamically hide show table selection column

Blakesalot:


packages/bits/src/lib/table/table.component.ts
```diff
@@ -221,14 +221,14 @@ export class TableComponent<T> extends CdkTable<T> implements OnInit, AfterViewI
...
         if (!ds) {
             this.dataSource = [];
         }
-        const changedDatasource = ds ? ds : [];
+        const changedDataSource = ds ? ds : [];
         // if no totalItems specified, we assume that there is only one page
-        this.tableStateHandlerService.changeDataSource(changedDatasource);
-        this.tableStateHandlerService.totalItems = this.totalItems || changedDatasource.length;
+        this.tableStateHandlerService.totalItems = this.totalItems || changedDataSource.length;
+        this.tableStateHandlerService.changeDataSource(changedDataSource);
```
Blakesalot, line 231: Switching these two lines fixes a minor bug where the selector had the "Select All Items on All Pages" option if there was only one page in the table right after page load. It didn't hurt anything, but this fixes it.

baggiosw (approved):

lukasholcik (approved):

PR #170, opened 2021-03-11
Feature/NUI 5752 continue work with snyk

Vita-Karas:


.circleci/config.yml
```diff
@@ -564,6 +586,10 @@ workflows:
                                 - main
                                 - /release\/.*/
                                 - next
+            -   snyk-test:
+                    requires:
+                        - dashboards-build
```
Blakesalot, line 591: It would be good to re-add the comment from the closed PR about waiting until the last package is built before running the snyk test.

Blakesalot (approved):

PR #171, opened 2021-03-11
percy config

dmytrofirsov:


Vita-Karas (approved):

.circleci/config.yml
```diff
@@ -472,6 +490,9 @@ workflows:
                         branches:
                             only:
                                 - main
+            - bits-percy-visual-test:
+                  requires:
+                      - bits-build
```
Blakesalot, line 495: Should we add the same branch filter as the one used by `bits-visual-test` starting at line 489?

packages/bits/.percy.yml
```diff
@@ -0,0 +1,29 @@
...
+  # (per-snapshot Percy CSS is concatenated to the end of this value)
+  # Details: https://docs.percy.io/docs/percy-specific-css
+  percyCSS: |
+    input, textarea { 
+      caret-color: transparent;
+    }
+    * {
+      cursor: none !important;
+    }
+  # WE DON'T CARE ABOUT THE CONFIGS BELOW
```
Blakesalot, line 18: Just curious, why are they here if we don't care about them? Maybe this information could be included in the comment?

Blakesalot (changes requested):

packages/bits/.percy.yml
dmytrofirsov: Theoretically, those options could be used for another way of taking snapshots - taking local screenshots and bulk-sending them to percy. I decided to leave them there for future reference, I think I'll be reviewing it in the next sprint in the scope of other percy tasks, so no worries.

.circleci/config.yml
dmytrofirsov: makes sense, let me change that

Blakesalot (approved):

PR #172, opened 2021-03-11
NUI-5715: Internal overlay doesn't disable in overlay component after…

NataliiaPodpriatova:
… main overlay is closed

### DIFF
--- a/packages/bits/demo/src/components/demo/overlay/overlay-custom-confirmation-inside-dialog/overlay-custom-confirmation-inside-dialog.component.ts
+++ b/packages/bits/demo/src/components/demo/overlay/overlay-custom-confirmation-inside-dialog/overlay-custom-confirmation-inside-dialog.component.ts
@@ -1,13 +1,17 @@
 import { FlexibleConnectedPositionStrategy, OverlayRef } from "@angular/cdk/overlay";
-import { Component, TemplateRef, ViewChild } from "@angular/core";
+import { Component, OnDestroy, TemplateRef } from "@angular/core";
 import { DialogService, NuiDialogRef, OverlayComponent } from "@nova-ui/bits";
+import { Subject } from "rxjs";
+import { takeUntil } from "rxjs/operators";
 
 @Component({
   selector: "nui-custom-confirmation-inside-dialog",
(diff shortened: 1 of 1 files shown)


packages/bits/demo/src/components/demo/overlay/overlay-custom-confirmation-inside-dialog/overlay-custom-confirmation-inside-dialog.component.ts
```diff
@@ -29,6 +33,9 @@ export class CustomConfirmationInsideDialogComponent {
         }]);
         // We update the size of the overlay container to follow the dimentions of the new 'toggle reference' container we set in the very first step
         this.updateOverlayDimensions(overlay);
+
+        // Handling ESC events inside overlay
+        this.activeDialog?.closed$.pipe(takeUntil(this.onDestroy$)).subscribe(() => overlay.hide());
```
Blakesalot, line 38: This will subscribe every time the overlay is triggered. So, it's a small memory leak.

This can be fixed by adding another takeUntil with a separate subject to the operator list:

```suggestion
        this.activeDialog?.closed$.pipe(
            takeUntil(this.overlayTriggered$)
            takeUntil(this.onDestroy$)
        ).subscribe(() => overlay.hide());
```

Then, at the top of the triggerOverlay, you can invoke `this.overlayTriggered$.next()` so that any previous subscription will be unsubscribed before you subscribe again.

For completeness, `this.overlayTriggered$.complete()` can be added to the ngOnDestroy as well.

Blakesalot (changes requested):

packages/bits/demo/src/components/demo/overlay/overlay-custom-confirmation-inside-dialog/overlay-custom-confirmation-inside-dialog.component.ts
NataliiaPodpriatova: Yes, I absolutely agree, it will be added

Blakesalot (approved):

PR #174, opened 2021-03-11
NUI-5784: Fix Roboto font in docs

madelyneh:
https://jira.solarwinds.com/browse/NUI-5784


packages/bits/angular.json
```diff
@@ -310,5 +311,8 @@
       "prefix": "nui",
       "lintFix": true
     }
+  },
+  "cli": {
+    "analytics": false
```
Blakesalot, line 316: I forget what this is for.

packages/bits/demo/src/static/typography/typography-example.component.html
```diff
@@ -5,7 +5,7 @@ <h4 class="mb-3">Core IT</h4>
 <nui-expander header="Use mixin">
     <ol>
         <li>
-            Import <strong>font-roboto.less</strong> directly like <mark>@import "@nova-ui/bits/sdk/less/mixins/font-roboto.less"</mark>
+            Import <strong>font-roboto.less</strong> directly like <mark>@import "@nova-ui/bits/dist/sdk/less/mixins/font-roboto.less"</mark>
```
Blakesalot, line 8: It doesn't look like the deployed bits package has a dist folder. It seems like this path should be able to stay as it was.

image

packages/bits/src/styles/nui-framework-variables.less
```diff
@@ -14,7 +14,7 @@
 // == Fonts
 @font-family-open-sans: "Open Sans", "Helvetica Neue", Helvetica, Arial, sans-serif;
 @font-family-source-code-pro: "Source Code Pro", monospace;
-@font-family-roboto: "Roboto", @font-family-open-sans;
+@font-family-roboto: "Roboto";
```
Blakesalot, line 17: It seems like we should keep the fallback font in case roboto doesn't get loaded for some reason.

packages/bits/demo/src/static/typography/typography-example.component.html
Blakesalot: So, we may need to look more into why consumers don't have access to the paths inside [this](https://github.com/solarwinds/nova/blob/3208133f89247f4bf4ca48c89c88e6958f7b26c8/packages/bits/src/styles/mixins/font-roboto.less#L6) file.

Blakesalot (changes requested):

packages/bits/angular.json
```diff
@@ -99,7 +99,8 @@
               "src/component-styles.less",
               "src/styles/nui-framework-theme.less",
               "src/styles/cdk-overlay-override.less",
-              "demo/src/demostyles.less"
+              "demo/src/demostyles.less",
+              "dist/bundles/css/styles-font-roboto.css"
```
Blakesalot, line 103: If consumers need to do this, do you mind updating the docs to include a code snippet at around line 21 of the docs page below?

Blakesalot (changes requested):
I think we should add a visual test to Dashboards to regression test that this works for consumers.

packages/bits/angular.json
madelyneh: This stops angular from asking you everytime you run the app if you want to send analytics back to to them. I can remove it if you like though

Blakesalot: seems fine to keep it :)

packages/bits/angular.json
madelyneh: Not sure what docs page you are meaning?

Blakesalot (approved):

PR #176, opened 2021-03-16
Bugfix/NUI-5866 timeseries widget empty data set domain

baggiosw:


lukasholcik (approved):

packages/charts/examples/components/demo/chart-types/line/line-chart-with-2y-axes/line-chart-with-2y-axes-example.component.ts
```diff
@@ -94,5 +94,10 @@ function getData() {
                 { x: moment("2017-01-03T15:14:29.909Z", format), y: 90 },
             ],
         },
+        {
+            id: "series-4",
+            name: "No data",
+            data: [],
+        },
```
Blakesalot, line 101: @baggiosw, it's not a super big deal, but ideally we should avoid changing examples for the sole purpose of verifying a feature or defect. We should reserve examples for explaining features and avoid adding any extra stuff that may distract from the feature being demonstrated. An alternative means to verify this might be to create a quick proof of concept somewhere in the development routes of the example app.

PR #180, opened 2021-03-18
NUI-5575 nui-sorter does not reflect direction input change

madelyneh:
https://jira.solarwinds.com/browse/NUI-5575


packages/bits/demo/src/components/demo/sorter/sorter-test/sorter-test.example.component.ts
```diff
@@ -0,0 +1,104 @@
+import { Component, OnInit, ViewChild } from "@angular/core";
```
madelyneh, line 1: These files dont need to be added. I just kept them in order to make verification earlier.

baggiosw (approved):

PR #191, opened 2021-03-24
feature/NUI-5750-Migrate-nova-charts-to-eslint

Vita-Karas:


packages/charts/examples/.eslintrc.json
```diff
@@ -0,0 +1,27 @@
+{
+  "extends": "../.eslintrc.json",
+  "ignorePatterns": [
+    "!**/*"
```
Blakesalot, line 4: Just out of curiosity what is this for?

Blakesalot (dismissed):

packages/charts/examples/.eslintrc.json
Vita-Karas: It is the default value for configured .eslintrc.json file. Will delete it so as not to confuse anyone :)

Blakesalot (approved):

PR #195, opened 2021-03-26
NUI-5791 FF only - The bottom thresholds label of the radial gauge is covered by the donut segment

Blakesalot:


madelyneh (approved):
So complicated. 🤯 Lol

PR #199, opened 2021-03-31
feature/NUI-5928-Add-aria-role-tags-to-components-Part-1

Vita-Karas:


packages/bits/src/lib/content/content.component.ts
```diff
@@ -10,6 +10,7 @@ import {Component, Input, ViewEncapsulation} from "@angular/core";
     templateUrl: "./content.component.html",
     styleUrls: ["./content.component.less"],
     encapsulation: ViewEncapsulation.None,
+    host: { "role": "contentinfo" },
```
Blakesalot, line 13: I'm not sure if contentinfo is the right role for this. The documentation for contentinfo says that it's used to identify information repeated at the end of every page of a website such as copyright info and navigation links.

image

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listbox_role

packages/bits/src/lib/date-picker/date-picker-inner.component.html
```diff
@@ -1,3 +1,3 @@
-<div *ngIf="datepickerMode" class="nui-datepicker-inner" role="application">
+<div *ngIf="datepickerMode" class="nui-datepicker-inner" role="dialog">
```
Blakesalot, line 1: I don't think `dialog` can be used here because date-picker-inner can either be inline or inside an overlay. Maybe date-picker-inner can have an input to specify the correct role?

https://github.com/solarwinds/nova/blob/main/packages/bits/src/lib/date-picker/date-picker.component.html#L28

packages/bits/src/lib/expander/expander.component.html
```diff
@@ -1,6 +1,7 @@
 <div class="nui-expander"
      [class.nui-expander--expanded]="state === 'expanded'">
     <div class="nui-expander__header"
+         role="button"
```
Blakesalot, line 4: `switch` might be even better since this button is a toggle.

packages/bits/src/lib/expander/expander.component.html
```diff
@@ -24,6 +25,7 @@
         </div>
     </div>
     <div class="nui-expander__body"
+         role="list"
```
Blakesalot, line 28: `list` may be two specific since the content inside the expander isn't necessarily a list of things.

packages/bits/src/lib/chips/chip/chip.component.ts
```diff
@@ -21,7 +21,10 @@ import { IChipsItem } from "../public-api";
     templateUrl: "./chip.component.html",
     styleUrls: ["./chip.component.less"],
     encapsulation: ViewEncapsulation.None,
-    host: { class: "nui-chip" },
+    host: {
+        class: "nui-chip",
+        role: "option",
```
Blakesalot, line 26: I'm not sure if `option` is correct here. This implies that you can select a chip, but I don't think chips are typically selectable.

packages/bits/src/lib/chips/chips.component.ts
```diff
@@ -124,6 +125,8 @@ export class ChipsComponent implements OnInit, OnDestroy, OnChanges, AfterViewIn
 
     private destroy$: Subject<void> = new Subject();
 
+    get role(): string | null { return this.getItemsCount() ? "listbox": null; }
```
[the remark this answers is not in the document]
Vita-Karas: will change to list

packages/bits/src/lib/content/content.component.ts
Vita-Karas: will change to listitem, thanks

packages/bits/src/lib/chips/chip/chip.component.ts
Vita-Karas: will change to listitem, thanks

packages/bits/src/lib/expander/expander.component.html
Vita-Karas: agree. Will add, thanks

packages/bits/src/lib/date-picker/date-picker-inner.component.html
Vita-Karas: You are right! will add, thanks

packages/bits/src/lib/expander/expander.component.html
Vita-Karas: will change to section

packages/bits/src/lib/date-picker/date-picker-inner.component.ts
```diff
@@ -75,6 +75,8 @@ export class DatePickerInnerComponent implements AfterContentInit, OnInit, OnCha
 
     private modes: string[] = ["day", "month", "year"];
 
+    get role(): string { return this.inline ? "application" : "dialog"; }
```
Blakesalot, line 78: I'm not sure that `application` applies here. The MDN web docs say it should only be used to define very dynamic, desktop-like web applications.

image
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Application_Role

packages/bits/src/lib/expander/expander.component.html
```diff
@@ -25,7 +25,7 @@
         </div>
     </div>
     <div class="nui-expander__body"
-         role="list"
+         role="section"
```
Blakesalot, line 28: Is this a valid role? I couldn't find it on the MDN site. I'm not sure what role would be appropriate here. Maybe `document`?

packages/bits/src/lib/date-picker/date-picker-inner.component.ts
Vita-Karas: I used this resource: https://www.w3.org/TR/wai-aria-1.1/#application as a single source of truth

packages/bits/src/lib/expander/expander.component.html
Vita-Karas: image

from https://www.w3.org/TR/wai-aria-1.1/

Blakesalot: Oh, I see. It looks like it's an abstract role, though, indicating that it shouldn't be used directly: https://www.w3.org/TR/wai-aria-1.1/#section

This page has the following:
image

packages/bits/src/lib/date-picker/date-picker-inner.component.ts
Blakesalot: okay

packages/bits/src/lib/expander/expander.component.html
Vita-Karas: I think for expander-body the role is unnecessary. Will removed.

Blakesalot (approved):

ivankozlovskyi (approved):

NataliiaPodpriatova (approved):

packages/bits/src/lib/expander/expander.component.html
```diff
@@ -1,6 +1,7 @@
 <div class="nui-expander"
      [class.nui-expander--expanded]="state === 'expanded'">
     <div class="nui-expander__header"
+         role="switch"
```
dmytrofirsov, line 4: The closes widget to the expander is Disclosure. For the disclodure type of elements, we have a set of requirements
https://www.w3.org/TR/wai-aria-practices-1.1/#disclosure

See the documentation for the details

ivankozlovskyi (approved):

dmytrofirsov (approved):

PR #202, opened 2021-04-01
NUI-5565: Tab navigation is not working properly in a form

NataliiaPodpriatova:


packages/bits/spec/components/textbox-number/textbox-number.atom.ts
```diff
@@ -40,4 +39,6 @@ export class TextboxNumberAtom extends Atom {
 
     public isValid = async (): Promise<boolean> => !(await this.rootElement.getAttribute("class")).includes("has-error");
 
+    public getInputId = async (): Promise<string> => await this.input.getId();
```
Blakesalot, line 42: It might be better to just make the `input` property public so that consumers can access the input's ElementFinder api directly. That way, we don't have to expand the API of this atom every time the consumer needs to access something else on the input element, and this `getInputId` method can just be removed.

packages/bits/spec/components/textbox-number/textbox-number.e2e.ts
```diff
@@ -171,4 +172,38 @@ describe("USERCONTROL textbox-number >", () => {
             expect(await component.getValue()).toBe("1");
         });
     });
+
+    describe("keyboard navigation", () => {
```
Blakesalot, line 176: Nice tests!

packages/bits/spec/components/textbox-number/textbox-number.atom.ts
NataliiaPodpriatova: Yes, you are right, it's a good point. Thank you

packages/bits/spec/components/textbox-number/textbox-number.e2e.ts
NataliiaPodpriatova: Thank you!

Blakesalot (dismissed):

Vita-Karas (dismissed):

packages/bits/spec/components/textbox-number/textbox-number.e2e.ts
```diff
@@ -171,4 +172,38 @@ describe("USERCONTROL textbox-number >", () => {
...
+        beforeEach(async () => {
+            await browser.refresh();
+            component = basic;
+        })
+
+        it("should navigate between elements using TAB", async () => {
+            expect("body")
+                .toEqual(await browser.switchTo().activeElement().getTagName());
```
dmytrofirsov, line 184: Shouldn't it be vise versa? :) Like, we usually expect the element tag to equal some string. Just curious

dmytrofirsov (approved):

packages/bits/spec/components/textbox-number/textbox-number.e2e.ts
NataliiaPodpriatova: yes, it's a good point, thank you. I'll fix it

Vita-Karas (approved):

dmytrofirsov (approved):

PR #205, opened 2021-04-02
feature/NUI-5929-Add-aria-role-tags-to-components-Part-2

Vita-Karas:


packages/bits/src/lib/convenience/time-frame-bar/time-frame-bar.component.ts
```diff
@@ -41,6 +41,7 @@ import { TimeFramePickerComponent } from "../../time-frame-picker/time-frame-pic
     templateUrl: "./time-frame-bar.component.html",
     styleUrls: ["./time-frame-bar.component.less"],
     changeDetection: ChangeDetectionStrategy.OnPush,
+    host: { "role": "navigation" },
```
Blakesalot, line 44: This is another role that doesn't quite fit the use case here. I'm not sure what it should be, but `navigation` doesn't seem right. `navigation` is supposed to be a group of the main links for a website.
image
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Navigation_Role

packages/bits/src/lib/menu/menu-popup/menu-popup.component.ts
```diff
@@ -27,6 +27,7 @@ import { IMenuGroup, IMenuItem } from "../public-api";
     selector: "nui-menu-popup",
     host: {
         "class": "nui-menu-popup",
+        "role": "dialog",
```
Blakesalot, line 30: it seems like `menu` might be more appropriate here. `dialog` is a little more generic and doesn't tell the user exactly what to expect from this component. what do you think?

packages/bits/src/lib/popover/popover-modal.component.ts
```diff
@@ -79,6 +80,8 @@ export class PopoverModalComponent implements AfterViewInit, OnInit, OnDestroy {
 
     private popoverModalSubscriptions: Subscription[] = [];
 
+    get role(): string { return this.backdrop ? "alertdialog" : "dialog"; }
```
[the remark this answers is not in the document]
Vita-Karas: I think, that it can be applicable for popover-modal.
image

https://www.w3.org/TR/wai-aria-1.1/#alertdialog

Also we can see the example here: https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/alertdialog.html

https://dequeuniversity.com/library/aria/simple-alert-dialog

packages/bits/src/lib/repeat/repeat-item/repeat-item.component.ts
```diff
@@ -21,4 +22,8 @@ export class RepeatItemComponent {
     @Input() public nowrap = false;
 
     @Input() public selected = false;
+
+    @Input() public draggable = false;
+
+    get role(): string { return this.clickable || this.draggable ? "option" : "listitem"; }
```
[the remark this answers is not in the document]
Vita-Karas: I was wrong here, because of using "option" for draggable elements. Will fixed.

But we have enough information about the clickable and selectable items. 
image
 It seems like one condition isItClickable will be enought for us.

packages/bits/src/lib/menu/menu-popup/menu-popup.component.ts
Vita-Karas: Agree. Will changed, thanks

packages/bits/src/lib/paginator/paginator.component.ts
```diff
@@ -46,6 +46,7 @@ const containerPaddingsWithScroll = 37;
     styleUrls: ["./paginator.component.less"],
     encapsulation: ViewEncapsulation.None,
     providers: [PopupContainerService],
+    host: { "role": "navigation" },
```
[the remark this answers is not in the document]
Vita-Karas: image

from https://a11y-style-guide.com/style-guide/section-navigation.html#kssref-navigation-pagination

we can use role="navigation" and will describe in aria-label which type of navigation are using here.

packages/bits/src/lib/convenience/time-frame-bar/time-frame-bar.component.ts
Vita-Karas: Maybe, spinbutton role be more applicable here

packages/bits/src/lib/repeat/repeat-item/repeat-item.component.ts
```diff
@@ -23,7 +23,5 @@ export class RepeatItemComponent {
 
     @Input() public selected = false;
 
-    @Input() public draggable = false;
-
-    get role(): string { return this.clickable || this.draggable ? "option" : "listitem"; }
+    get role(): string { return this.clickable ? "option" : "listitem"; }
```
Blakesalot, line 26: I'm still not sure that `clickable` provides enough information to know whether the repeat item is a selectable `option` because it could be clickable in the sense that you can click it to navigate somewhere or do some other action besides selecting the item. The correct role may need to be an input whose value comes from the repeat component.

packages/bits/src/lib/selector/selector.component.html
```diff
@@ -25,7 +24,6 @@
 <div #popupArea></div>
 
 <nui-overlay #overlay
-             [role]="'menu'"
              [toggleReference]="toggleButton"
```
Blakesalot: should the roleAttr be specified here?

packages/bits/src/lib/sorter/sorter.component.html
```diff
@@ -22,7 +21,6 @@
     <div #popupArea></div>
 
     <nui-overlay #overlay
-                 [role]="'menu'"
                  [toggleReference]="toggleRef"
```
Blakesalot: should the roleAttr be specified here?

packages/bits/src/lib/selector/selector.component.html
Vita-Karas: No, because of `nui-menu-popup` inside `nui-overlay`. `nui-menu-popup` under the hood has the role="menu"

packages/bits/src/lib/sorter/sorter.component.html
Vita-Karas: No, because of `nui-menu-popup` inside `nui-overlay`. `nui-menu-popup` under the hood has the role="menu"

packages/bits/src/lib/repeat/repeat-item/repeat-item.component.ts
Vita-Karas: Discussed offline and added `selectable` input for that purpose

Blakesalot (approved):

ivankozlovskyi (approved):

NataliiaPodpriatova (approved):

dmytrofirsov (approved):

PR #206, opened 2021-04-05
axe scanner lib upgrade

dmytrofirsov:


packages/bits/package.json
```diff
@@ -138,12 +138,12 @@
     "@angular/platform-browser-dynamic": "^11.1.2",
     "@angular/router": "^11.1.2",
     "@applitools/eyes-protractor": "1.9.1",
+    "@axe-core/webdriverjs": "4.1.1",
```
Blakesalot, line 141: should we add the `^` to the beginning so that we can automatically receive minor updates like security fixes?

packages/bits/spec/components/form-field/form-field.a11y.ts
```diff
@@ -28,73 +28,74 @@ describe("a11y: form-field", () => {
     });
 
     it("button", async () => {
-        const accessibilityScanResults = await axeBuilder(browser.driver).include(`.${ButtonAtom.CSS_CLASS}`).disableRules("color-contrast").analyze();
+        const accessibilityScanResults = await new AxeBuilder(browser.driver).include(`.${ButtonAtom.CSS_CLASS}`).disableRules("color-contrast").analyze();
```
Blakesalot, line 31: can we new this up in a beforeAll to reuse the same instance every time, instead of creating a new one for each test?

dmytrofirsov: Unfortunately not. If we'll use a single instance of axe it'll preserve the results of any previously run test, which will result in false negatives.

Creating a new instance is fine according to their docs:
https://github.com/dequelabs/axe-core-npm/tree/develop/packages/webdriverjs

packages/bits/package.json
dmytrofirsov: good idea, will add in a sec

Blakesalot (approved):

NataliiaPodpriatova (approved):

PR #212, opened 2021-04-08
NUI-5946: Text area scrollable - default behavior

dmytrognatyshyn:


packages/bits/demo/src/components/demo/textbox/textbox-visual-test/textbox-visual-test.component.ts
```diff
@@ -8,6 +8,9 @@ import { Component } from "@angular/core";
 export class TextboxVisualTestComponent {
     public isRequired = true;
     public value = "";
+    public textExample: string = 'Lorem ipsum dolor sit amet, consectetur adipisicing elit. Aliquid dolores ' +
+        'ex harum itaque laborum laudantium quae quo, soluta suscipit voluptate. Accusamus aliquid amet architecto, ' +
+        'doloremque ducimus eveniet facere facilis fugit itaque';
```
dmytrofirsov, line 13: You can use backticks for multiline strings, so you don't have to concatenate them.

packages/bits/demo/src/components/demo/textbox/textbox-visual-test/textbox-visual-test.component.html
```diff
@@ -99,5 +99,17 @@ <h3>Textbox with Textarea and Placeholder</h3>
             </nui-textbox>
             <hr />
         </div>
+        <div class="w-100"></div>
+        <div class="col">
+            <h3>Textbox with scrollable Textarea</h3>
```
Blakesalot, line 104: It might be good to also have an automation test for non-scrollable to ensure that consumers can still turn it off if necessary.

Blakesalot:
It seems like something went wrong with a merge somehow because this PR is now showing 104 changed files.

PR #215, opened 2021-04-09
NUI-5984  Automatically adjust donut gauge thickness based on donut size

Blakesalot:


packages/charts/src/renderers/radial/gauge/donut-gauge-renderer-config.ts
```diff
@@ -7,7 +7,6 @@ import { IRadialRendererConfig } from "../../../core/common/types";
  */
 export function donutGaugeRendererConfig(): IRadialRendererConfig {
     return {
-        annularGrowth: 0,
```
Blakesalot, line 10: @elenates, the annularGrowth override of zero has been removed for the gauge renderer so the gauge thickness will grow at the same rate as a standard donut chart.

packages/charts/src/core/grid/radial-grid.ts
```diff
@@ -15,15 +16,19 @@ export class RadialGrid extends Grid implements IGrid {
...
-        this.config().dimension
-            .outerWidth(dimensions.width)
-            .outerHeight(dimensions.height);
+        const dimensionConfig = this.config().dimension;
+        if (!isUndefined(dimensions.width)) {
+            dimensionConfig.outerWidth(dimensions.width)
+        }
+        if (!isUndefined(dimensions.height)) {
+            dimensionConfig.outerHeight(dimensions.height)
+        }
```
Blakesalot, line 25: There was a bug with the radial grid where you couldn't manually set the grid dimensions without a crash here.

elenates (approved):

lukasholcik (approved):

PR #216, opened 2021-04-13
NUI-5946 Text area scrollable default behavior

dmytrognatyshyn:


Blakesalot (approved):

dmytrognatyshyn:
It may potentially be breaking. Moved to v12

PR #218, opened 2021-04-13
NUI-5901 Added alt for images

dmytrognatyshyn:


packages/bits/src/lib/image/image.component.ts
```diff
@@ -116,8 +126,9 @@ export class ImageComponent implements OnInit, AfterViewInit {
         let imageHtml: string = "";
         if (_has(image, "code") && _isString(image.code)) {
             imageHtml = image.code;
+            this.ariaLabel = image.name;
```
Blakesalot, line 129: It looks like `ariaLabel` will only be populated if `image` has a `code` property. What do you think about putting this outside of the if check and doing something like this:

```
this.ariaLabel = this.imageAlt || image.name;
```

That way we would default to using the `imageAlt` as the aria label and fall back to the `image.name` if no `imageAlt` is provided. It's just an idea.

packages/bits/src/lib/image/image.component.ts
```diff
@@ -116,8 +126,11 @@ export class ImageComponent implements OnInit, AfterViewInit {
         let imageHtml: string = "";
         if (_has(image, "code") && _isString(image.code)) {
             imageHtml = image.code;
+            if(image.name && this.ariaLabel !== image.name) {
+                Promise.resolve().then(_ => this.ariaLabel = image.name);
+            }
         } else {
-            imageHtml = `<img src="${this.image}">`;
+            imageHtml = `<img src="${this.image}" alt="${this.imageAlt}">`;
```
Vita-Karas, line 133: Please, add the default `imageAlt`. If the user does not add `imageAlt`, the screen reader will read the image src.

packages/bits/src/lib/image/image.component.ts
```diff
@@ -66,6 +74,8 @@ export class ImageComponent implements OnInit, AfterViewInit {
      */
     @Input() public autoFill: boolean;
 
+    public ariaLabel: string;
```
Vita-Karas, line 77: The `ariaLabel` need to be an input. It gives the ability for the user to set their own `ariaLabel`.

packages/bits/src/lib/image/image.component.ts
dmytrognatyshyn: I had the same idea in the beginning but when I did it, screen readers read 2 times the same text. First, it reads from  and then from component aria-label. So I decided to left only alt. 
I'll add input ariaLabel to the component. So users will have possibility to set their own aria-label.

packages/bits/src/lib/image/image.component.ts
dmytrognatyshyn: I have added empty string as default value for imageAlt.  If the user does not add imageAlt, the screen reader will read only the component role ('image') and aria-label (if it will be added)

packages/bits/src/lib/image/image.component.ts
```diff
@@ -116,8 +131,11 @@ export class ImageComponent implements OnInit, AfterViewInit {
         let imageHtml: string = "";
         if (_has(image, "code") && _isString(image.code)) {
             imageHtml = image.code;
+            if(image.name && !this.ariaLabel && this.imageName !== image.name) {
+                Promise.resolve().then(_ => this.imageName = image.name);
```
lukasholcik, line 135: Setting an internal property from a getter function is not a good practice, especially when the function is called from the template. That's why you had to wrap it in a promise. I'd prefer to find another way around that.

Overall this function seems quite complex to be called from a template, so updating it wouldn't be a bad thing to do.

DmytroPutrin (dismissed):

packages/bits/src/lib/image/image.component.ts
```diff
@@ -5,8 +5,8 @@ import {
     Component,
     ElementRef,
     Inject,
-    Input,
-    OnInit,
+    Input, OnChanges,
+    OnInit, SimpleChanges,
```
Blakesalot, line 9: newlines needed here

packages/bits/src/lib/image/image.component.ts
```diff
@@ -99,6 +100,13 @@ export class ImageComponent implements OnInit, AfterViewInit {
         });
     }
 
+    ngOnChanges(changes: SimpleChanges): void {
```
Blakesalot, line 103: would be good to have an explicit `public` access modifier here

Blakesalot (approved):

packages/bits/src/lib/image/image.component.ts
```diff
@@ -5,8 +5,10 @@ import {
     Component,
     ElementRef,
     Inject,
-    Input,
-    OnInit,
+    Input, 
+    OnChanges,
+    OnInit,
```
Blakesalot, line 10: I think lines 8 and 10 have a trailing space which is causing the build to fail here due to lint errors: https://app.circleci.com/pipelines/github/solarwinds/nova/1690/workflows/f8485c15-1382-4984-a2f4-0bd1e2f57708/jobs/13742

Blakesalot (approved):

packages/bits/src/lib/image/image.component.ts
dmytrognatyshyn: Thank you. I already found it

Vita-Karas (approved):

packages/bits/src/lib/image/image.component.ts
dmytrofirsov: It looks like we've ended up here with two inputs that are doing the same thing.

I think we can easily throw the `ariaLabel` input to the `alt` attribute depending on the conditionals. We could also remove `aria-label` in case `alt` is used and vise versa to prevent screen readers from reading the same info twice.

dmytrofirsov (approved):

dmytrofirsov (approved):

DmytroPutrin (approved):

Vita-Karas (approved):

ivankozlovskyi (approved):

PR #219, opened 2021-04-13
Bugfix/n UI 5504 charts popover tooltip has width issue

baggiosw:


packages/bits/src/lib/popover/popover.component.ts
```diff
@@ -446,6 +457,7 @@ export class PopoverComponent implements OnDestroy, OnInit, OnChanges {
             .flexibleConnectedTo(this.host.nativeElement)
             .withPush(false)
             .withViewportMargin(0)
+            .withGrowAfterOpen(true)
```
baggiosw, line 460: Sets whether the overlay can grow after the initial open via flexible width/height.
--

packages/bits/src/lib/popover/popover.component.ts
```diff
@@ -285,6 +285,17 @@ export class PopoverComponent implements OnDestroy, OnInit, OnChanges {
         this.overlayComponent?.getOverlayRef()?.updatePosition();
     }
 
+    /**
+     * Resets the size of the popover based on the overlay's position.
```
Blakesalot, line 289: How about the following since there isn't anything about position in the implementation of this method?

```suggestion
     * Resets the size of the popover.
```

Blakesalot (approved):

elenates (approved):

PR #227, opened 2021-04-16
Bugfix/n UI 3143 dialog is not dismissed upon navigation change with browser buttons

dmytrognatyshyn:


Blakesalot (approved):

packages/bits/spec/components/dialog/dialog.e2e.ts
```diff
@@ -138,6 +140,23 @@ describe("USERCONTROL Dialog", () => {
             expect(await dialog.isDialogDisplayed()).toBe(true);
         });
     });
+
+    describe("Dismissal on route changes", async () => {
+        afterAll(async ()=> {
+           await browser.navigate().back();
+        })
+
+        it("should close dialog with router changed", async () => {
```
Blakesalot, line 149: nice test 👍

DmytroPutrin (dismissed):

ivankozlovskyi (dismissed):

packages/bits/spec/components/dialog/dialog.e2e.ts
```diff
@@ -28,6 +28,8 @@ describe("USERCONTROL Dialog", () => {
     let cancelButton: ElementFinder;
     let actionButton: ElementFinder;
     let themeSwitcher: WebElement;
+    const setLocation = (url: string) =>
+        browser.executeScript((pUrl: string) => window.location.href = `/#/${pUrl}`, url);
```
dmytrofirsov, line 32: `browser.executeScript()` returns a Promise, so the function itself must be async. Later in the code, when calling the function, we'll need to await it to properly resolve the returned promise.

dmytrofirsov:
very nice job overall!

dmytrofirsov (dismissed):

packages/bits/spec/components/dialog/dialog.e2e.ts
Blakesalot: @dmytrofirsov, do you think this would be a handy function to put in the Helpers class along with the existing `prepareBrowser` function? It's just an idea and not really necessary for this PR.

dmytrofirsov: Makes sense @Blakesalot . @dmytrognatyshyn  please add this function to the helpers class, make it static and 
 describe what the function does in a comment

dmytrognatyshyn: @dmytrofirsov ok

dmytrofirsov (approved):

DmytroPutrin (approved):

ivankozlovskyi (approved):

PR #228, opened 2021-04-19
NUI-5440: Timeseries legend click anchor styling

madelyneh:
NUI-5440-timeseries-legend-click-anchor-styling


packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.less
```diff
@@ -0,0 +1,5 @@
+@import (reference) "node_modules/@nova-ui/bits/sdk/less/nui-framework-variables";
+
+.description-primary {
```
lukasholcik, line 3: Is this a nova-charts class? If yes, then I'd prefer adding our own class with a BEM compatible name to avoid styling a generic class.

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.ts
```diff
@@ -70,6 +70,13 @@ export abstract class XYChartComponent extends TimeseriesChartComponent
         return this.configuration.legendPlacement === LegendPlacement.Right;
     }
 
+    /** Checks if the series is interactive or if the dataSource has a link property */
+    public interactionOrLink(legendSeries: IChartAssistSeries<IAccessors>): boolean {
```
lukasholcik, line 74: I think a better name would be "isInteractive" or "isSeriesInteractive".

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.html
```diff
@@ -26,6 +26,41 @@
                                       [color]="chartAssist.palette.standardColors.get(legendSeries.id)">
                     </nui-chart-marker>
                 </nui-rich-legend-tile>
+
+                <div description class="d-flex align-items-center px-2">
+                    <div class="description-container d-flex flex-column justify-content-center">
+                        <div *ngIf="(legendSeries?.link? false : true) else elseLinkHref">
```
Blakesalot, line 32: Could be simplified in this way:

```suggestion
                        <div *ngIf="!legendSeries?.link else elseLinkHref">
```

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.html
```diff
@@ -26,6 +26,41 @@
...
+                                    [title]="legendSeries.legendDescriptionSecondary">
+                                        {{legendSeries.legendDescriptionSecondary}}
+                                </div>
+                            </div>
+                        </div>
+                        <ng-template #elseLinkHref>
+                            <div (click)="$event.stopPropagation()">
+                                <a class="nui-text-link-small"
+                                        [href]="legendSeries.link" 
+                                        target="_blank">
```
Blakesalot, line 49: I don't know if we should use target `_blank` here. I think users might appreciate being able to CTRL click or right click to open in a new tab if they want. If they want to open the new page in the same tab, they should be able to just click it. Perhaps this is another UX question for Thursday?

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.ts
```diff
@@ -70,6 +70,13 @@ export abstract class XYChartComponent extends TimeseriesChartComponent
         return this.configuration.legendPlacement === LegendPlacement.Right;
     }
 
+    /** Checks if the series is interactive or if the dataSource has a link property */
+    public isSeriesInteractive(legendSeries: IChartAssistSeries<IAccessors>): boolean {
+        const link: boolean = !(legendSeries?.link === undefined) || !(this.seriesInteractive === undefined);
+        return link;
+
```
Blakesalot, line 77: extra newline

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.ts
```diff
@@ -70,6 +70,13 @@ export abstract class XYChartComponent extends TimeseriesChartComponent
         return this.configuration.legendPlacement === LegendPlacement.Right;
     }
 
+    /** Checks if the series is interactive or if the dataSource has a link property */
+    public isSeriesInteractive(legendSeries: IChartAssistSeries<IAccessors>): boolean {
+        const link: boolean = !(legendSeries?.link === undefined) || !(this.seriesInteractive === undefined);
+        return link;
```
Blakesalot, line 76: could be simplified to the following:

```suggestion
        return !(legendSeries?.link === undefined) || !(this.seriesInteractive === undefined);
```

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.ts
```diff
@@ -85,6 +92,23 @@ export abstract class XYChartComponent extends TimeseriesChartComponent
...
+        } else {
+            this.eventBus.getStream(INTERACTION).next({
+                payload: {
+                    data: legendSeries,
+                    interactionType: TimeseriesInteractionType.Series,
+                },
+            });
+        }
+    }
+
```
Blakesalot, line 111: unit tests needed for this new method

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.ts
```diff
@@ -70,6 +70,13 @@ export abstract class XYChartComponent extends TimeseriesChartComponent
         return this.configuration.legendPlacement === LegendPlacement.Right;
     }
 
+    /** Checks if the series is interactive or if the dataSource has a link property */
+    public isSeriesInteractive(legendSeries: IChartAssistSeries<IAccessors>): boolean {
```
Blakesalot, line 74: unit tests needed for this method as well

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.html
madelyneh: That makes complete sense. Removed for now.

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.html
```diff
@@ -26,6 +26,41 @@
...
+                            <div class= "nui-text-link-small"
+                                (click)="onInteraction(legendSeries)">
+                                <div class="description description-primary legend-description-primary" 
+                                    [title]="legendSeries.legendDescriptionPrimary">
+                                        {{legendSeries.legendDescriptionPrimary}}
+                                </div>
+                                <div class="description description-secondary" 
+                                    [title]="legendSeries.legendDescriptionSecondary">
+                                        {{legendSeries.legendDescriptionSecondary}}
+                                </div>
```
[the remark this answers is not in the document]
madelyneh: I believe the elements are already getting those styles when they are injected into the legend? That is why I used those classes from charts.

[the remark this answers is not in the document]
Blakesalot: If they are, I'm wondering why you needed to add a less file for the font.

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.html
```diff
@@ -4,28 +4,78 @@
...
-                               [descriptionPrimary]="legendSeries.legendDescriptionPrimary"
-                               [descriptionSecondary]="legendSeries.legendDescriptionSecondary"
-                               [isSelected]="!chartAssist.isSeriesHidden(legendSeries.id)"
-                               (isSelectedChange)="chartAssist.toggleSeries(legendSeries.id, $event)"
-                               (mouseenter)="chartAssist.emphasizeSeries(legendSeries.id)"
-                               (mouseleave)="chartAssist.resetVisibleSeries()"
-                               [seriesRenderState]="chartAssist.renderStatesIndex[legendSeries.id]?.state"
-                               (click)="onLegendClick(legendSeries, $event)">
+                        [descriptionPrimary]="undefined"
+                        [descriptionSecondary]="undefined"
```
Blakesalot, line 14: Can we just delete these instead of setting them to undefined?

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.html
```diff
@@ -4,28 +4,78 @@
...
+                                [marker]="chartAssist.markers.get(legendSeries.id)"
+                                [color]="chartAssist.palette.standardColors.get(legendSeries.id)">
                     </nui-chart-marker>
                 </nui-rich-legend-tile>
+
+                <div description
+                    class="legend-description-container nui-text-small d-flex flex-column justify-content-center px-2"
+                    [ngClass]="['state-' + chartAssist.renderStatesIndex[legendSeries.id]?.state]">
+
+                        State - {{chartAssist.renderStatesIndex[legendSeries.id]?.state}}
```
Blakesalot, line 34: I'm assuming these are here only temporarily :)

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.html
```diff
@@ -4,28 +4,78 @@
...
+                            [title]="legendSeries.legendDescriptionPrimary">
+                                HREF - {{legendSeries.legendDescriptionPrimary}}
+                        </a>
+                        <div (click)="onInteraction(legendSeries, $event)"
+                            class="legend-description-secondary"
+                            *ngIf="!legendSeries.secondaryLink" 
+                            [title]="legendSeries.legendDescriptionSecondary">
+                                {{legendSeries.legendDescriptionSecondary}}
+                        </div>
+                        <div *ngIf="legendSeries.secondaryLink">
```
Blakesalot, line 57: is this extra div necessary? I notice that the <a> tag on line 44 doesn't have a surrounding div.

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.html
```diff
@@ -4,28 +4,78 @@
...
+                            <a class="legend-description-secondary link"
+                            [href]="legendSeries.secondaryLink" 
+                            rel="noopener noreferrer"
+                            [title]="legendSeries.legendDescriptionSecondary">
+                            HREF - {{legendSeries.legendDescriptionSecondary}}
```
Blakesalot, line 62: indentation

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.less
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        &-secondary {
+            .setCssVariable(color, nui-color-text-secondary);
+            margin-top: -1px;
+            &.link {
+                .setCssVariable(color, nui-color-text-link);
+                &:hover {
+                    text-decoration: underline;
+                }
```
Blakesalot, line 40: I think we might be able to use the typography provided in bits for this: https://github.com/solarwinds/nova/blob/main/packages/bits/src/styles/nui-framework-typography.less#L223-L230

Blakesalot: @import (reference) "@nova-ui/bits/sdk/less/nui-framework-typography.less";

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.html
```diff
@@ -4,28 +4,78 @@
...
+                            <div *ngIf="!!legendSeries.legendDescriptionPrimary" 
+                                class="legend-description-primary" 
+                                [title]="legendSeries.legendDescriptionPrimary">{{ legendSeries.legendDescriptionPrimary }}
+
+                            </div>
+                            <div *ngIf="!!legendSeries.legendDescriptionSecondary" class="legend-description-secondary" [title]="legendSeries.legendDescriptionSecondary">{{ legendSeries.legendDescriptionSecondary }}
+
+                            </div>
+                        </div>
+                    </ng-template>
```
Blakesalot, line 77: I wonder if we actually need this extra template here. It seems like it would be possible to use ngClass on the other templates above to conditionally add the `link` class which would provide the correct styling and hover behavior for links only when necessary.

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.less
```diff
@@ -0,0 +1,52 @@
...
+            }
+        }
+    }
+    &:not(:hover)&.state-hidden {
+        .legend-description {
+            opacity: @nui-legend-deemphasized-opacity;
+            .legend-description-primary {
+                color: @nui-color-text-default;
```
Blakesalot, line 48: is setCssVariable needed here? It seems like it should be used here

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.less
```diff
@@ -0,0 +1,52 @@
...
+
+@nui-legend-deemphasized-opacity: 0.3;
+@nui-legend-series-content-margin: 3px;
+
+.legend-description-container {
+    & .legend-description {
+        // line-height: @nui-line-height-default;
+        white-space: nowrap;
+        text-overflow: ellipsis;
+        overflow: hidden;
```
Blakesalot, line 13: The mixin `.text-overflow(ellipsis)` could be used to replace these lines. For reference, the mixin is found here: https://github.com/solarwinds/nova/blob/main/packages/bits/src/styles/mixins/text-overflow.less

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.spec.ts
```diff
@@ -166,5 +166,23 @@ describe("XYChartComponent", () => {
...
             expect(spy).toHaveBeenCalledWith({ payload: timeframe });
         });
+
+        it("should return true if the series is set to interactive", () => {
+            component.configuration = { interaction: "series" } as ITimeseriesWidgetConfig;
+            component.ngOnChanges(initializationChanges);
+            expect(component.isSeriesInteractive(component.chartAssist.legendSeriesSet[0])).toBe(true);
+        });
+
+        it("should communitcate interaction on the EventBus when series is interactive", () => {
```
Blakesalot, line 176:
```suggestion
        it("should communicate interaction on the EventBus when series is interactive", () => {
```

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.ts
```diff
@@ -70,6 +70,11 @@ export abstract class XYChartComponent extends TimeseriesChartComponent
...
+        return !(legendSeries?.link === undefined) || !(legendSeries?.secondaryLink === undefined) || !(this.seriesInteractive === undefined);
```
Blakesalot, line 75: `this.seriesInteractive` is a boolean, so we should use the value as is instead of checking for undefined.

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.ts
```diff
@@ -85,6 +90,28 @@ export abstract class XYChartComponent extends TimeseriesChartComponent
...
+        }
+        const target = event.target as HTMLElement;
+        const secondaryDescription = target.classList.contains("legend-description-secondary");
+
+        if (legendSeries.link && !secondaryDescription) {
+            this.eventBus.getStream(INTERACTION).next({ payload: { data: legendSeries } });
+        } else if (secondaryDescription) {
+            this.chartAssist.toggleSeries(legendSeries.id, this.chartAssist.isSeriesHidden(legendSeries.id));
+        }
+    }
```
Blakesalot, line 113: Would it be possible to separate this logic out into two separate click handlers? One for `onPrimaryDescClick` and one for `onSecondaryDescClick`. That may help to simplify the logic here, as you wouldn't have to query the event for the target class list.

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.ts
```diff
@@ -85,6 +90,28 @@ export abstract class XYChartComponent extends TimeseriesChartComponent
         }
     }
 
+    public onInteraction(legendSeries: IChartAssistSeries<IAccessors>, event: MouseEvent) {
+        let interactive = this.isSeriesInteractive(legendSeries);
+        if (!interactive) {
+            console.log("📘 xy-chart.component: 96# -> legendSeries:", legendSeries);
```
Blakesalot, line 96: debug log can be removed. you're probably already planning on doing that :)

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.ts
```diff
@@ -85,6 +90,28 @@ export abstract class XYChartComponent extends TimeseriesChartComponent
         }
     }
 
+    public onInteraction(legendSeries: IChartAssistSeries<IAccessors>, event: MouseEvent) {
+        let interactive = this.isSeriesInteractive(legendSeries);
```
Blakesalot, line 94: The result of this query may be too broad. What if there's a secondary link and no primary link, and the primary description is clicked? I think the event would be sent regardless. But, in that case, I think we should avoid emitting the event.

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.spec.ts
```diff
@@ -166,5 +166,23 @@ describe("XYChartComponent", () => {
             });
             expect(spy).toHaveBeenCalledWith({ payload: timeframe });
         });
+
+        it("should return true if the series is set to interactive", () => {
+            component.configuration = { interaction: "series" } as ITimeseriesWidgetConfig;
+            component.ngOnChanges(initializationChanges);
+            expect(component.isSeriesInteractive(component.chartAssist.legendSeriesSet[0])).toBe(true);
```
Blakesalot, line 173: We may need tests for `link` and `secondaryLink` as well since if those are truthy, this method should return. Additionally, if they're all false, this method should return false.

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.spec.ts
```diff
@@ -166,5 +166,23 @@ describe("XYChartComponent", () => {
...
+            const spy = spyOn(eventBus.getStream(INTERACTION), "next").and.callThrough();
+            component.configuration = { interaction: "series" } as ITimeseriesWidgetConfig;
+            component.ngOnChanges(initializationChanges);
+            component.onInteraction(component.chartAssist.legendSeriesSet[0]);
+            const interactionData = {
+                payload: { data: component.chartAssist.legendSeriesSet[0], interactionType: TimeseriesInteractionType.Series },
+                id: "INTERACTION",
+            };
+            expect(spy).toHaveBeenCalledWith(interactionData);
+        });
```
Blakesalot, line 186: I don't thing the interaction event should be fired if it's a simple `<a>` tag. Correct @lukasholcik? If not, we should probably have tests that check to make sure the event isn't emitted in case the primary and/or secondary descriptions are just simple links with `<a>` tags.

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.html
madelyneh: I am honestly not sure. But when it isn't surrounded by the div tag then it will not be displayed at all.

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.html
madelyneh: If there was a way to conditionally add a click handler to the tag as well I would agree with you. But i don't believe there is?

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.html
```diff
@@ -4,28 +4,77 @@
...
-                               [descriptionPrimary]="legendSeries.legendDescriptionPrimary"
-                               [descriptionSecondary]="legendSeries.legendDescriptionSecondary"
-                               [isSelected]="!chartAssist.isSeriesHidden(legendSeries.id)"
-                               (isSelectedChange)="chartAssist.toggleSeries(legendSeries.id, $event)"
-                               (mouseenter)="chartAssist.emphasizeSeries(legendSeries.id)"
-                               (mouseleave)="chartAssist.resetVisibleSeries()"
-                               [seriesRenderState]="chartAssist.renderStatesIndex[legendSeries.id]?.state"
-                               (click)="onLegendClick(legendSeries, $event)">
+            <nui-legend-series class="legend-container"
+                        [ngClass]="['state-' + chartAssist.renderStatesIndex[legendSeries.id]?.state]"
```
Blakesalot, line 13: it seems like this should already be taken care of if you use the seriesRenderState input:
https://github.com/solarwinds/nova/blob/0a504b1c5c55cafc5bd02e2845e6f0da1791f591/packages/charts/src/legend/legend-series/legend-series.component.html#L2

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.html
```diff
@@ -4,28 +4,77 @@
...
-            <nui-legend-series *ngFor="let legendSeries of chartAssist.legendSeriesSet; trackBy: chartAssist.seriesTrackByFn"
-                               [descriptionPrimary]="legendSeries.legendDescriptionPrimary"
-                               [descriptionSecondary]="legendSeries.legendDescriptionSecondary"
-                               [isSelected]="!chartAssist.isSeriesHidden(legendSeries.id)"
-                               (isSelectedChange)="chartAssist.toggleSeries(legendSeries.id, $event)"
-                               (mouseenter)="chartAssist.emphasizeSeries(legendSeries.id)"
-                               (mouseleave)="chartAssist.resetVisibleSeries()"
-                               [seriesRenderState]="chartAssist.renderStatesIndex[legendSeries.id]?.state"
-                               (click)="onLegendClick(legendSeries, $event)">
+            <nui-legend-series class="legend-container"
```
Blakesalot, line 12: is this class used anywhere? I can't find a usage.

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.html
```diff
@@ -4,28 +4,77 @@
         <nui-chart class="w-100" [chart]="chartAssist.chart"></nui-chart>
     </div>
     <div class="d-flex" *ngIf="hasLegend()"
-         [ngClass]="[legendShouldBeAlignedRight() ? 'legend scrollable' : 'legend-bottom']">
+                [ngClass]="[legendShouldBeAlignedRight() ? 'legend scrollable' : 'legend-bottom']">
```
Blakesalot, line 7: indentation

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.less
```diff
@@ -0,0 +1,48 @@
...
+                }
+            }
+        }
+    }
+
+    &:not(:hover)&.state-hidden {
+        .legend-description-container {
+            opacity: @nui-legend-deemphasized-opacity;
+            .legend-description-primary {
+                .setCssVariable(color, nui-color-text-default);
```
Blakesalot, line 41: Do we know whether we're supposed to remove the link styling for hidden series? I would think if you can still click it to navigate somewhere, we should keep the link styling, no?

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.less
```diff
@@ -0,0 +1,48 @@
...
+                    .setCssVariable(color, nui-color-text-secondary);
+                }
+                &:not(:first-child) {
+                    margin-top: -1px;
+                }
+            }
+        }
+    }
+
+    &:not(:hover)&.state-hidden {
```
Blakesalot, line 37: this class doesn't look correct to me. here's what it looks like in the nui-legend-series less file:

https://github.com/solarwinds/nova/blob/0a504b1c5c55cafc5bd02e2845e6f0da1791f591/packages/charts/src/legend/legend-series/legend-series.component.less#L140

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.html
Blakesalot: I think you should be able to do something like this:
```html
(click)="isSeriesInteractive ? onClick() : undefined"
```

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.spec.ts
```diff
@@ -166,5 +166,61 @@ describe("XYChartComponent", () => {
...
+            component.widgetData.series[0] = {
+                id: "id",
+                name: "name",
+                description: "description",
+                secondaryLink: "www.link.com",
+                data: {} as ITimeseriesWidgetSeriesData[],
+            }
+            component.ngOnChanges(initializationChanges);
+            expect(component.isSeriesInteractive(component.chartAssist.legendSeriesSet[0])).toBe(true);
+        });
```
Blakesalot, line 200: Do you mind putting these three tests in their own describe block for `isSeriesInteractive` outside of the `ngOnChanges` describe block.

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.spec.ts
```diff
@@ -166,5 +166,61 @@ describe("XYChartComponent", () => {
...
+            component.widgetData = { series: [{} as ITimeseriesWidgetData] };
+            component.widgetData.series[0] = {
+                id: "id",
+                name: "name",
+                description: "description",
+                data: {} as ITimeseriesWidgetSeriesData[],
+            }
+            component.ngOnChanges(initializationChanges);
+            expect(component.isSeriesInteractive(component.chartAssist.legendSeriesSet[0])).toBe(false);
+        });
```
Blakesalot, line 224: This one can also go in the new `isSeriesInteractive` describe block.

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.spec.ts
Blakesalot: It looks like we still need tests to ensure that the event isn't emitted when it's a simple link for both the primary and secondary descriptions. And, maybe a test for ensuring that it's not emitted when there's neither a simple link nor a data source-provided interaction.

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.ts
```diff
@@ -70,6 +70,11 @@ export abstract class XYChartComponent extends TimeseriesChartComponent
         return this.configuration.legendPlacement === LegendPlacement.Right;
     }
 
+    /** Checks if the series is interactive or if the dataSource has a link property */
+    public isSeriesInteractive(legendSeries: IChartAssistSeries<IAccessors>): boolean | undefined {
+        return !(legendSeries?.link === undefined) || !(legendSeries?.secondaryLink === undefined) || !!this.seriesInteractive;
```
Blakesalot, line 75: Since this.seriesInteractive is just a boolean, the double bang isn't necessary. (No need to convert a boolean to a boolean)

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.html
Blakesalot: On second thought, it's probably better to go ahead and set the click handler, and just filter out clicks in the handler implementation with an if check:
```ts
public onClick() {
    if (this.isSeriesInteractive) {
        // ...do interactive stuff
    }
}
```

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.ts
```diff
@@ -85,6 +90,19 @@ export abstract class XYChartComponent extends TimeseriesChartComponent
...
+        this.eventBus.getStream(INTERACTION).next({
+            payload: {
+                data: legendSeries,
+                interactionType: TimeseriesInteractionType.Series,
+            },
+        });
+    }
+
+    public onSecondaryDescClick(legendSeries: IChartAssistSeries<IAccessors>) {
+        this.chartAssist.toggleSeries(legendSeries.id, this.chartAssist.isSeriesHidden(legendSeries.id));
```
Blakesalot, line 103: It seem like if we conditionally remove the `stopPropagation` from the secondaryDescription element in the case where there's no link that the nui-legend-series should handle the series toggling for you. Can you give that a try instead?

packages/charts/src/legend/legend-series/legend-series.component.ts
```diff
@@ -16,8 +16,6 @@ export const LEGEND_SERIES_CLASS_NAME = "nui-legend-series";
     host: { "class": LEGEND_SERIES_CLASS_NAME },
     selector: "nui-legend-series",
     templateUrl: "./legend-series.component.html",
-    styleUrls: ["./legend-series.component.less"],
```
Blakesalot, line 19: The file itself should be deleted tool

packages/charts/src/nui-charts.less
```diff
@@ -1,5 +1,9 @@
 @import (reference) "@nova-ui/bits/sdk/less/nui-framework-variables";
 @import (reference) "nui-charts-variables.less";
+@import (reference) "@nova-ui/bits/sdk/less/mixins.less";
+
+@nui-legend-series-max-width: 250px;
+@nui-legend-series-description-min-width: 130px; // 180px from mockups - 50px for
```
Blakesalot, line 6:
```suggestion
@nui-legend-series-description-min-width: 130px; // 180px from mockups - 50px for tile
```

packages/charts/src/nui-charts.less
```diff
@@ -73,3 +77,151 @@
...
+            opacity: @nui-legend-deemphasized-opacity;
+        }
+    }
+
+    & &--state-deemphasized {
+        opacity: @nui-legend-deemphasized-opacity;
+    }
+}
+
+.description-min-width--unset {
```
Blakesalot, line 223: It seems like this one should go at line 107 just above `.description` since the name isn't really specific enough to be out here on its own. And, once its there, to be safe I wonder if we should add an `!important` next to the rule to ensure that it takes priority. I found the actual usage of this class here:
https://github.com/solarwinds/nova/blob/0c3bd4215aa7054e2488b02de9acf19921bd7bde/packages/dashboards/src/lib/components/proportional-widget/proportional-widget.component.html#L48

Blakesalot: it will need a `&` in front of it with no space of course

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.html
```diff
@@ -4,28 +4,61 @@
...
-                               [descriptionPrimary]="legendSeries.legendDescriptionPrimary"
-                               [descriptionSecondary]="legendSeries.legendDescriptionSecondary"
-                               [isSelected]="!chartAssist.isSeriesHidden(legendSeries.id)"
-                               (isSelectedChange)="chartAssist.toggleSeries(legendSeries.id, $event)"
-                               (mouseenter)="chartAssist.emphasizeSeries(legendSeries.id)"
-                               (mouseleave)="chartAssist.resetVisibleSeries()"
-                               [seriesRenderState]="chartAssist.renderStatesIndex[legendSeries.id]?.state"
-                               (click)="onLegendClick(legendSeries, $event)">
+            <nui-legend-series 
+                    [ngClass]="'legend-container '+ 'state-' + chartAssist.renderStatesIndex[legendSeries.id]?.state"
```
Blakesalot, line 13: Do you mind trying to use the `[seriesRenderState]` input now that we've made the styles public? The nui-legend-series should add the correct render state classes based on the chart assist's renderStatesIndex.

Here's a usage of that:
https://github.com/solarwinds/nova/blob/0c3bd4215aa7054e2488b02de9acf19921bd7bde/packages/charts/examples/components/demo/advanced-usage/legend/legend-interactive/legend-interactive.example.component.html#L19

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.less
Blakesalot: If we use the `[seriesRenderState]` input on the nui-legend-series, I think we can use the classes that are automatically applied to the series instead of this custom class `state-hidden`.

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.less
```diff
@@ -0,0 +1,19 @@
+@import (reference) "@nova-ui/bits/sdk/less/nui-framework-variables.less";
+@import (reference) "@nova-ui/bits/sdk/less/mixins.less";
+
+.legend-container {
```
Blakesalot, line 4: I think we can use the existing `.nui-legend-series` class here instead. This class is automatically applied to the nui-legend-series element.

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.html
```diff
@@ -4,28 +4,61 @@
...
                     <nui-chart-marker rightEdge
-                                      [marker]="chartAssist.markers.get(legendSeries.id)"
-                                      [color]="chartAssist.palette.standardColors.get(legendSeries.id)">
+                                [marker]="chartAssist.markers.get(legendSeries.id)"
+                                [color]="chartAssist.palette.standardColors.get(legendSeries.id)">
                     </nui-chart-marker>
                 </nui-rich-legend-tile>
+
+                <div description class="description-container d-flex flex-column nui-text-small justify-content-center">
+
```
Blakesalot, line 31: extra newline

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.html
```diff
@@ -4,28 +4,61 @@
...
+                        <div *ngIf="!legendSeries.link" 
+                            class="description description-primary" 
+                            [ngClass]="{'nui-text-link-small link': isSeriesInteractive(legendSeries.id)}"
```
Blakesalot, line 34: I think the result of isSeriesInteractive may be too vague to know whether these link classes should be applied to the primary description. For example, if there's a secondary link but no primary link, it will return true and add the link styling to the primary description even thought there's no link for the primary.

Blakesalot: Maybe we should used the `this.seriesInteractive` property instead since that tells us whether there's a urlInteractionHandler?

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.ts
```diff
@@ -70,6 +70,11 @@ export abstract class XYChartComponent extends TimeseriesChartComponent
         return this.configuration.legendPlacement === LegendPlacement.Right;
     }
 
+    /** Checks if the series is interactive or if the dataSource has a link property */
+    public isSeriesInteractive(legendSeries: IChartAssistSeries<IAccessors>): boolean {
+        return !(legendSeries?.link === undefined) || !(legendSeries?.secondaryLink === undefined) || this.seriesInteractive;
+    }
+
     public onLegendClick(legendSeries: IChartAssistSeries<IAccessors>, event: MouseEvent) {
```
Blakesalot, line 78: It seems like this method should be removed now that we're handling the interaction emission with onPrimaryDescClick

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.ts
```diff
@@ -85,6 +90,23 @@ export abstract class XYChartComponent extends TimeseriesChartComponent
         }
     }
 
+    public onPrimaryDescClick(legendSeries: IChartAssistSeries<IAccessors>) {
+        if (!this.seriesInteractive) {
+            this.chartAssist.toggleSeries(legendSeries.id, this.chartAssist.isSeriesHidden(legendSeries.id));
```
Blakesalot, line 95: This shouldn't be necessary if you change it to conditionally `stopPropagation` only if `this.seriesInteractive`. Currently, `stopPropagation` is invoked every time the legend is clicked.

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.ts
Blakesalot: See previous comment. I still think this is the case.

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.spec.ts
```diff
@@ -167,4 +167,71 @@ describe("XYChartComponent", () => {
...
+            const spy = spyOn(eventBus.getStream(INTERACTION), "next").and.callThrough();
+            component.configuration = { interaction: "series" } as ITimeseriesWidgetConfig;
+            component.ngOnChanges(initializationChanges);
+            component.onPrimaryDescClick(component.chartAssist.legendSeriesSet[0]);
+            const interactionData = {
+                payload: { data: component.chartAssist.legendSeriesSet[0], interactionType: TimeseriesInteractionType.Series },
+                id: "INTERACTION",
+            };
+            expect(spy).toHaveBeenCalledWith(interactionData);
+        });
```
Blakesalot, line 223: This test seems out of place in the isSeriesInteractive describe block. Maybe it deserves an onPrimaryDescClick describe block of its own?

Blakesalot: And, maybe add a test for the following description as well:
```
it("should not communicate interaction on the EventBus when series is not interactive", () => {
```

Blakesalot:
I almost forgot. We should have a visual test for this.

packages/charts/src/nui-charts.less
```diff
@@ -73,3 +77,151 @@
...
+        margin-right: @nui-space-sm;
+    }
+
+    &:empty {
+        display: none;
+    }
+
+    .description-min-width--unset {
+        .description {
+            min-width: unset!important;
```
Blakesalot, line 110: space missing between unset and !important

packages/charts/src/nui-charts.less
```diff
@@ -73,3 +77,151 @@
...
+
+    &--horizontal:not(:last-child) {
+        margin-right: @nui-space-sm;
+    }
+
+    &:empty {
+        display: none;
+    }
+
+    .description-min-width--unset {
```
Blakesalot, line 108: I think we need an ampersand here since the class is added directly to the nui-legend-series element in dashboards (a.k.a not to a child):

https://github.com/solarwinds/nova/blob/0c3bd4215aa7054e2488b02de9acf19921bd7bde/packages/dashboards/src/lib/components/proportional-widget/proportional-widget.component.html#L48

```suggestion
    &.description-min-width--unset {
```

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.html
```diff
@@ -4,28 +4,57 @@
...
-                                      [marker]="chartAssist.markers.get(legendSeries.id)"
-                                      [color]="chartAssist.palette.standardColors.get(legendSeries.id)">
+                                [marker]="chartAssist.markers.get(legendSeries.id)"
+                                [color]="chartAssist.palette.standardColors.get(legendSeries.id)">
                     </nui-chart-marker>
                 </nui-rich-legend-tile>
+                <div description class="description-container d-flex flex-column nui-text-small justify-content-center">
+                        <div *ngIf="!legendSeries.link" 
+                            class="description description-primary" 
+                            [ngClass]="{'nui-text-link-small link': ((legendSeries?.link) || seriesInteractive) }"
```
Blakesalot, line 32: I just realized we don't need the check for `legendSeries?.link` here because `legendSeries?link` is always false for this element.

```suggestion
                            [ngClass]="{'nui-text-link-small link': seriesInteractive }"
```

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.ts
```diff
@@ -70,19 +70,18 @@ export abstract class XYChartComponent extends TimeseriesChartComponent
         return this.configuration.legendPlacement === LegendPlacement.Right;
     }
 
-    public onLegendClick(legendSeries: IChartAssistSeries<IAccessors>, event: MouseEvent) {
+    public onPrimaryDescClick(event: any, legendSeries: IChartAssistSeries<IAccessors>) {
```
Blakesalot, line 73: It's better to put a type here if we can. This will likely make your unit test need an `as MouseEvent` after the argument value.

```suggestion
    public onPrimaryDescClick(event: MouseEvent, legendSeries: IChartAssistSeries<IAccessors>) {
```

Blakesalot (approved):

elenates (approved):

elenates (approved):

PR #230, opened 2021-04-19
Bugfix/n NUI-5993 Dynamically adding a step does not copy nextText from html

dmytrognatyshyn:


packages/bits/src/lib/wizard/wizard.component.ts
```diff
@@ -156,8 +156,15 @@ export class WizardComponent implements OnInit, AfterContentInit, AfterViewCheck
         const componentFactory = this.componentFactoryResolver.resolveComponentFactory(WizardStepComponent);
         const componentRef = this.dynamicStep.createComponent(componentFactory);
         const instance: IWizardStepComponent = componentRef.instance;
-        instance.title = wizardStep.title;
-        instance.stepTemplate = wizardStep.stepTemplate;
+        const wizardStepInputs = [
```
lukasholcik, line 159: Doing this has all sorts of side effects and requires maintenance when new properties are added. I'd reconsider the approach here. If you want, we can have a call about this.

lukasholcik (changes requested):

DmytroPutrin (approved):

Vita-Karas (approved):

packages/bits/src/lib/wizard/wizard.component.ts
dmytrognatyshyn: @lukasholcik I made some changes. Please take a look

PR #231, opened 2021-04-19
Research/n UI 5921 nova docs refactor

baggiosw:


.circleci/config.yml
```diff
@@ -666,6 +666,10 @@ workflows:
             -   nova-docs:
                     context:
                         - common-build-creds
+                    filters:
+                        branches:
+                            ignore:
+                                - pull/*
```
baggiosw, line 672: This is for ignoring forked pull request

.circleci/config.yml
```diff
@@ -510,11 +510,18 @@ jobs:
...
+        docker:
+            -   image: circleci/node:12
+        environment:
+            NOVA_DIR: ~/nova
+        steps:
             -   run:
                     name: Docs Cleanup
                     command: yarn run clean-docs-ci
-            -   store_artifacts:
-                    path: ./nova-docs.zip
```
Blakesalot, line 517: do we need to store the cleaned up version of the docs in the artifacts?

.circleci/config.yml
Blakesalot: It seems like we should publish docs for forked PRs too if possible to make it easier to QA external contributions before merge.

.circleci/config.yml
```diff
@@ -700,6 +711,10 @@ workflows:
             -   snyk-test:
                     requires:
                         - dashboards-build
+                    filters:
+                        branches:
+                            ignore:
+                                - pull/*
```
Blakesalot, line 717: It seems like we should run snyk tests on forked PR branches as well.

scripts/deploy-docs
```diff
@@ -66,4 +66,5 @@ git config user.name "${GITHUB_USER_NAME}"
 git remote set-url origin https://$GITHUB_USER_NAME:$GITHUB_TOKEN@github.com/solarwinds/nova-docs.git
 git add .
 git commit -m "add docs" --allow-empty
+git pull -X theirs
```
Blakesalot, line 69: what was this for exactly?

.circleci/config.yml
baggiosw: We don't pass tokens so that might be hard to do

scripts/deploy-docs
baggiosw: This will pull any changes and if there are any merge conflicts uses theirs.  It will fix some issues with that error we see.

.circleci/config.yml
baggiosw: I disagree this can be a very bad backdoor at worst it will create on main

.circleci/config.yml
baggiosw: Github review made this look bad

.circleci/config.yml
ewok-janitor: I agree with baggio- security first, even if it complicates forks

.circleci/config.yml
Blakesalot: okay, at least any new vulnerabilities would theoretically be caught in main.

.circleci/config.yml
lukasholcik: Yeah, you need github credentials via environment var to make this succeed. Forked PRs don't have access to that.

ewok-janitor: if we changed the snyk scanner to not use tokens (and just block PRs), this might work

.circleci/config.yml
ewok-janitor: If we changed the snyk scanner to just block PRs (and not use tokens), this might work.

elenates (approved):

PR #234, opened 2021-04-19
NUI-5236: Nested expander body overlaps next html elements

madelyneh:
https://jira.solarwinds.com/browse/NUI-5236

### DIFF
--- a/packages/bits/src/animations/expand.ts
+++ b/packages/bits/src/animations/expand.ts
@@ -3,7 +3,9 @@ import {
     state,
     style,
     transition,
-    trigger
+    trigger,
+    stagger,
+    query
 } from "@angular/animations";
 
 export const expand = trigger("expandedState", [
@@ -12,4 +14,22 @@ export const expand = trigger("expandedState", [
     transition("expanded <=> collapsed", [
         animate("350ms ease-in-out"),
     ]),
+]);
+
+    /**
+     * This animation is using the special selectors that ngIf and ngFor use on newly inserted or removed content.
+     */
+export const expandV2 = trigger('expandedAddition', [
+    transition(":enter", [
+        style({ height: 0 }),
+        animate("350ms ease-in", style({ height: "*" })),
+        query("@*", 
+            stagger(300, [
+                animate("350ms ease-in", style({ height: "*" })),
+            ]),{optional: true})
+    ]),
+    transition(":leave", [
+        style({ height: "*" }),
+        animate("350ms ease-out", style({ height: 0 })),
+    ])
 ]);
--- a/packages/bits/src/animations/expand.ts
+++ b/packages/bits/src/animations/expand.ts
@@ -32,7 +32,7 @@ export const expandV2 = trigger("expandContent", [
                 style({ height: 0}),
                 stagger(10, [
                     animate("250ms ease-in-out", style({ height: "*" })),
-                ])
+                ]),
             ], {optional: true}),
         ]),
     ]),
--- a/packages/bits/src/animations/expand.ts
+++ b/packages/bits/src/animations/expand.ts
@@ -19,17 +19,17 @@ export const expand = trigger("expandedState", [
     /**
      * This animation is using the special selectors that ngIf and ngFor use on newly inserted or removed content.
      */
-export const expandV2 = trigger('expandedAddition', [
+export const expandV2 = trigger("expandContent", [
     transition(":enter", [
         style({ height: 0 }),
         animate("350ms ease-in", style({ height: "*" })),
-        query("@*", 
+        query("@*",
             stagger(300, [
                 animate("350ms ease-in", style({ height: "*" })),
-            ]),{optional: true})
+            ]),{optional: true}),
     ]),
     transition(":leave", [
         style({ height: "*" }),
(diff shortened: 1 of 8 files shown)


packages/bits/src/lib/expander/expander.component.html
```diff
@@ -29,6 +29,9 @@
          [class.hide-left-border]="hideLeftBorder"
          [@expandedState]="state"
          (@expandedState.done)="expandedStateDone()">
-        <ng-content *ngIf="renderBody"></ng-content>
+            <div *ngIf="renderBody else noBody">
+                <ng-content></ng-content>
+            </div>
+            <ng-template #noBody></ng-template>
```
Blakesalot, line 35: is this empty template necessary?

packages/bits/src/lib/expander/expander.component.html
```diff
@@ -29,6 +29,9 @@
          [class.hide-left-border]="hideLeftBorder"
          [@expandedState]="state"
          (@expandedState.done)="expandedStateDone()">
-        <ng-content *ngIf="renderBody"></ng-content>
+            <div *ngIf="renderBody else noBody">
```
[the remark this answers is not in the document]
Blakesalot: Also, an inline comment about why this extra div is necessary may be helpful.

Blakesalot (changes requested):
See comments for requested changes

packages/bits/src/animations/expand.ts
```diff
@@ -13,3 +15,21 @@ export const expand = trigger("expandedState", [
         animate("350ms ease-in-out"),
     ]),
 ]);
+
+    /**
+     * This animation is using the special selectors that ngIf and ngFor use on newly inserted or removed content.
+     */
```
Blakesalot, line 21: Indentation can be removed. Also, I would add something about why this animation is better. For example, "This v2 version of the expand animation respects the animations of elements nested inside the expander's content template."

One more thing, do you think we should deprecate the old expand animation? I think we maybe should do that. If so, it would be good to add a jsdoc comment like the following. And, then of course create the ticket for removal of the old animation to reference in the comment.
```
/** @deprecated As of Nova v11, use 'expandV2' instead. Removal: NUI-???? */
```

packages/bits/src/animations/expand.ts
```diff
@@ -13,3 +15,21 @@ export const expand = trigger("expandedState", [
...
+export const expandV2 = trigger("expandContent", [
+    transition(":enter", [
+        style({ height: 0 }),
+        animate("350ms ease-in", style({ height: "*" })),
+        query("@*",
+            stagger(300, [
```
Blakesalot, line 27: Does this need to be >= to 350 (instead of 300) to meet or exceed the 350ms animation time? Is that possibly why there is a jump when expanding the nested ones?

packages/bits/src/lib/expander/expander.component.html
```diff
@@ -26,9 +26,9 @@
         </div>
     </div>
     <div class="nui-expander__body"
-         [class.hide-left-border]="hideLeftBorder"
-         [@expandedState]="state"
-         (@expandedState.done)="expandedStateDone()">
-        <ng-content *ngIf="renderBody"></ng-content>
+        [class.hide-left-border]="hideLeftBorder"
+         *ngIf="open"
+        @expandContent>
```
Blakesalot, line 31: indentation alignment issue

packages/bits/src/lib/expander/expander.component.html
```diff
@@ -26,9 +26,9 @@
         </div>
     </div>
     <div class="nui-expander__body"
-         [class.hide-left-border]="hideLeftBorder"
-         [@expandedState]="state"
-         (@expandedState.done)="expandedStateDone()">
-        <ng-content *ngIf="renderBody"></ng-content>
```
Blakesalot, line 32: I noticed that the `renderBody` property still exists in the component. Do we need to keep it around, or can we get rid of it?

packages/bits/src/lib/expander/expander.component.html
```diff
@@ -26,9 +26,9 @@
         </div>
     </div>
     <div class="nui-expander__body"
-         [class.hide-left-border]="hideLeftBorder"
-         [@expandedState]="state"
```
Blakesalot, line 30: I wonder if we can get rid of the `state` property in the component and just get away with a simple boolean value for `open` instead of the complex setter and getter methods for `open` that we currently have.

packages/bits/src/animations/expand.ts
madelyneh: The stagger is just a delay for each element to enter the animation. We dont want the stagger to be the same as the animation time because then it wont start the animation until the parent finishes it's animation. Does that make sense?

packages/bits/src/lib/expander/expander.component.html
madelyneh: We still need the state and that logic in order to populate the class with expanded or collapsed.

Blakesalot: okay

baggiosw (dismissed):

Blakesalot (approved):

baggiosw (approved):

PR #235, opened 2021-04-20
Feature/n UI 4140 wizard remove step

DmytroPutrin:


Blakesalot:
@dmytrofirsov, @DmytroPutrin, and @ewok-janitor, this PR adds the functionality to the legacy wizard. I just want to make sure that we indeed decided to fix the old wizard as opposed to fixing the new wizard for IPAM. If we're fixing the old, should we have a follow up ticket to verify the same functionality on the new wizard?

ewok-janitor:
> @dmytrofirsov, @DmytroPutrin, and @ewok-janitor, this PR adds the functionality to the legacy wizard. I just want to make sure that we indeed decided to fix the old wizard as opposed to fixing the new wizard for IPAM. If we're fixing the old, should we have a follow up ticket to verify the same functionality on the new wizard?

We have the existing wizard v2 epic.  That redesign should incorporate all the feedback (including consumer issues) we know about.

The reason for patching the old wizard is that IPAM REALLY wants these fixed, in a short amount of time.  A full wizard refactor is not possible in the timeframe (and staffing situation) we are in.  It is not my favorite situation, but its the best we can do at the moment.

PR #249, opened 2021-04-23
Feature/n UI 5769 write a11y tests for nova bits components

dmytrofirsov:


DmytroPutrin (approved):

packages/bits/spec/components/switch/switch.a11y.ts
```diff
@@ -0,0 +1,17 @@
...
+        "color-contrast", // NUI-6014
+    ];
+
+    beforeAll(async () => {
+        await Helpers.prepareBrowser("switch/switch-visual-test");
+    });
+
+    it("should check a11y of switch", async () => {
+        await assertA11y(browser, SwitchAtom.CSS_CLASS, rulesToDisable);
+    });
```
ivankozlovskyi, line 16: Maybe it would be good to add one more test with switch - on

ivankozlovskyi (commented):
astonishing work!

packages/bits/spec/components/switch/switch.a11y.ts
dmytrofirsov: the testing page itself has the switch component in "on" state

ivankozlovskyi (approved):

PR #256, opened 2021-04-27
NUI-5999: wizard-step.component.ts overwrite 'steps' class property f…

DmytroPutrin:
…or WizardDirective

### DIFF
--- a/packages/bits/src/lib/wizard-v2/wizard-step.component.ts
+++ b/packages/bits/src/lib/wizard-v2/wizard-step.component.ts
@@ -98,6 +98,7 @@ export class WizardDirective extends CdkStepper implements AfterContentInit {
 
     /** Steps that the stepper holds. */
     @ContentChildren(WizardStepV2Component, {descendants: true}) _steps: QueryList<WizardStepV2Component>;
+    @ContentChildren(WizardStepV2Component, {descendants: true}) steps: QueryList<WizardStepV2Component>;
 
     /** The step that is selected. */
     @Input()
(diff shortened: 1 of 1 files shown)


packages/bits/src/lib/wizard-v2/wizard-step.component.ts
```diff
@@ -98,6 +98,7 @@ export class WizardDirective extends CdkStepper implements AfterContentInit {
 
     /** Steps that the stepper holds. */
     @ContentChildren(WizardStepV2Component, {descendants: true}) _steps: QueryList<WizardStepV2Component>;
+    @ContentChildren(WizardStepV2Component, {descendants: true}) steps: QueryList<WizardStepV2Component>;
```
Blakesalot, line 101: Do we need two of these properties? `_steps` and `steps`

DmytroManco: yes. We wrap CDK Stepper component, that use both property and refs to CDK Step component. If we want to use our custom WizardStepV2Component, we should overwrite QueryList.
https://github.com/angular/components/blob/master/src/cdk/stepper/stepper.ts#L268 and line 271

Blakesalot: Oh, wait. It looks like this `steps` property is overriding the parent CdkStepper `steps` property. I'm not sure that this is a good thing.

Blakesalot: Okay, do you mind adding an inline comment about that to help prevent confusion in the future?

DmytroPutrin: Comment added.

ivankozlovskyi (approved):

dmytrognatyshyn (approved):

Vita-Karas (approved):

PR #273, opened 2021-05-04
bugfix/nui 6026 select v2 empty states

baggiosw:


packages/bits/src/lib/select-v2/base-select-v2.ts
```diff
@@ -338,7 +338,7 @@ export abstract class BaseSelectV2 implements AfterViewInit, AfterContentInit, C
             this._selectedOptions.forEach(option => option.outfiltered = true);
 
         } else {
-            const modelValue: OptionValueType = value;
+            const modelValue: OptionValueType | null = value;
```
Blakesalot, line 341: Should we maybe go ahead and add `null` to the `OptionValueType` itself?

packages/bits/src/lib/select-v2/base-select-v2.ts
```diff
@@ -318,11 +318,11 @@ export abstract class BaseSelectV2 implements AfterViewInit, AfterContentInit, C
     }
 
     protected getValueFromOptions(options = this.selectedOptions): OptionValueType | OptionValueType[] | null {
-        return this.multiselect ? options.map(o => o.value) : options[0]?.value || null;
+        return this.multiselect ? options.map(o => o.value) : options[0]?.value || "";
```
Blakesalot, line 321: What does this do?

packages/bits/src/lib/select-v2/base-select-v2.ts
```diff
@@ -387,7 +387,9 @@ export abstract class BaseSelectV2 implements AfterViewInit, AfterContentInit, C
     }
 
     private setActiveItemOnDropdown(): void {
-        this.value && !this.multiselect
+        const selectedValue = this.options?.find(option => isEqual(option.value, this.value));
```
Blakesalot, line 390: `this.value` may be one option or an array of options in the case of multi select. It seems like we need to handle multi select here as well.

packages/bits/src/lib/select-v2/select/select-v2.component.spec.ts
```diff
@@ -460,6 +460,14 @@ describe("components >", () => {
                     component.toggleDropdown();
                     expect(component["optionKeyControlService"].getActiveItemIndex()).toEqual(0);
                 });
+                it("should select the null option if value is null", () => {
+                    const nullOption = TestBed.createComponent(SelectV2OptionComponent).componentInstance
+                    component.selectedOptions = [nullOption];
```
Blakesalot, line 465: It may improve the test to add three options. Put the null one in the middle and check that the middle one is selected. This is because the first one gets selected in either case, so it could be a false positive.

packages/bits/src/lib/select-v2/select/select-v2.component.spec.ts
```diff
@@ -460,6 +460,14 @@ describe("components >", () => {
                     component.toggleDropdown();
                     expect(component["optionKeyControlService"].getActiveItemIndex()).toEqual(0);
                 });
+                it("should select the null option if value is null", () => {
```
Blakesalot, line 463: missing newline

packages/bits/src/lib/select-v2/base-select-v2.ts
baggiosw: It checks to see if the value is in the list of options and if it is not it will set it to "" instead of null.  This gets ran anytime selectedOptions gets ran.

Blakesalot (approved):

elenates (approved):

PR #275, opened 2021-05-05
bugfix/NUI-6022 Textbox number button focus outline doesn't perfectly align with button

dmytrognatyshyn:


packages/bits/src/lib/textbox/textbox-number/textbox-number.component.less
```diff
@@ -45,10 +46,12 @@
...
         top: 1px;
         text-align: center;
-        .nui & button.nui-button {
+        button.nui-button {
             padding-left: @nui-space-xs;
             width: (@nui-space-md * 2);
-            height: 28px;
+            height: @textbox-button-height;
+            min-height: @textbox-button-height;
+            padding-top: 7px;
```
Blakesalot, line 54: It would be good to move this to line 50 so that it's next to the other padding rule. Additionally, it would be good to use the variable `@button-padding-is-empty` which is equal to 6px and made available in buttons.less. Then, you can add 1px to that to account for the 1px height difference on the top due to overriding the default button height of 30px down to 28px.

The import for the variable would look something like:
```
@import (reference) "../../styles/components/buttons.less";
```

Blakesalot:
```suggestion
            // adding 1px to padding to account for adjusted button height
            padding-top: @button-padding-is-empty + 1px;
```

packages/bits/src/lib/textbox/textbox-number/textbox-number.component.less
```diff
@@ -7,6 +7,7 @@
 @textbox-padding-simple-updown: @textbox-updown-offset + @textbox-feedback-icon-margin;
 @textbox-feedback-icon-with-updown-margin: @textbox-feedback-icon-margin + @textbox-updown-offset;
 @textbox-busy-offset: 1px;
+@textbox-button-height: 28px;
```
Blakesalot, line 10: This can be improved by calculating the correct height of the buttons by subtracting the border (times 2) from input height base value.

Like this:

```suggestion
@textbox-button-height: @input-height-base - (@nui-line-default * 2);
```

Here's where the form control styling is set giving it the height via a direct rule and a border via the textbox-base() mixin:

https://github.com/solarwinds/nova/blob/891476fc40a072e0725ced68a786da2d9d143577/packages/bits/src/styles/components/forms.less#L35

dmytrognatyshyn: Fixed. Thank you @Blakesalot for the detailed description

DmytroPutrin (approved):

ivankozlovskyi (approved):

PR #276, opened 2021-05-05
Feature/NUI-5915 overflow on timeseries legend is too restrictive

Blakesalot:


packages/bits/demo/src/components/demo/common/unit-conversion-service/docs/unit-conversion-service-docs.component.less
```diff
@@ -1,12 +1,7 @@
-:host {
-    ul {
-        padding-left: 40px;
+ul {
```
Blakesalot, line 1: Feedback from previous PR

packages/bits/src/services/unit-conversion.service.ts
```diff
@@ -38,20 +36,13 @@ export class UnitConversionService {
...
-            if (valueCeiled % base === 0) {
-                resultValue = valueCeiled / base;
-                resultOrder += 1;
-            }
-
             strValue = (resultValue).toFixed(scale);
 
             // remove trailing zeros
-            strValue = parseFloat(strValue).toString();
+            strValue = parseFloat(strValue).toLocaleString(undefined, { maximumFractionDigits: scale });
```
Blakesalot, line 42: Added localization

packages/charts/src/core/common/scales/linear-scale.ts
```diff
@@ -9,7 +9,7 @@ export class LinearScale extends Scale<number> {
     constructor(id?: string) {
         super(id);
 
-        this.formatters.tick = value => value + "";
+        this.formatters.tick = value => value?.toLocaleString();
```
Blakesalot, line 12: linear scale tick labels are now localized by default

packages/dashboards/examples/src/components/app/prototype-index.component.ts
```diff
@@ -0,0 +1,69 @@
+import { Component } from "@angular/core";
```
Blakesalot, line 1: Component for easy access to prototypes and test pages similar to what we have for the charts demo app.

packages/dashboards/src/lib/components/timeseries-widget/types.ts
```diff
@@ -23,7 +23,7 @@ export interface ITimeseriesWidgetSeries {
 
 export interface ITimeseriesWidgetData<T = ITimeseriesWidgetSeriesData> {
     id: string;
-    name: string;
+    name?: string; // unused
```
Blakesalot, line 26: @lukasholcik, I set this property to optional because it doesn't appear to be used anywhere. If it's optional, consumers can use the interface without having to set the name unnecessarily.

packages/bits/src/services/unit-conversion.service.ts
```diff
@@ -38,20 +36,13 @@ export class UnitConversionService {
                 this.logger.warn("unit conversion service does not support conversion to negative order of magnitude");
             }
 
-            // fix the precision edge case
-            const valueCeiled = Math.ceil(resultValue);
-            if (valueCeiled % base === 0) {
-                resultValue = valueCeiled / base;
-                resultOrder += 1;
-            }
```
Blakesalot, line 46: There was a bug where values approaching conversion to the next higher unit would be prematurely converted. Removing this fixes that.

I pinpointed the origin of this block of code [here](https://github.com/solarwinds/orion-nova-novajs/commit/ecd43c058b50af9777f3e309b3d632e9efa9afac#diff-32871140f4e96e556288e082303b67f32901f119c54bd3ebb9d148d668afc248), and we've maintained the same [unit tests](https://github.com/solarwinds/orion-nova-novajs/commit/ecd43c058b50af9777f3e309b3d632e9efa9afac#diff-14de2330bc8407bdb155138ca489f254d72ef401d878d69fea889cc134f0d40f) that were created in the same PR as this code's origin.

All of our current unit tests pass, and I'm unable to detect what the "edge case" referred to in the comment might have been.

packages/bits/src/services/unit-conversion.service.spec.ts
```diff
@@ -388,12 +389,26 @@ describe("services >", () => {
...
                     expectedValue: "-1 kHz",
                 }, {
-                    name: "1000 Hz as 1 kHz with scale 4 and without tailing zeros",
+                    name: "1000 Hz as 1 kHz with scale 4 and without trailing zeros",
                     inputValue: 1000,
                     unit: "hertz",
                     scale: 4,
                     expectedValue: "1 kHz",
                 },
+                // Uncomment in the scope of NUI-6056
```
Blakesalot, line 398: https://jira.solarwinds.com/browse/NUI-6056

madelyneh (approved):

PR #277, opened 2021-05-06
[NUI-5845] select-v2, combobox-v2 fix double scroll in examples with …

DmytroPutrin:
…wirtual scroll


packages/bits/demo/src/components/demo/combobox-v2/combobox-v2-virtual-scroll/combobox-v2-virtual-scroll.example.component.html
```diff
@@ -1,4 +1,5 @@
 <nui-combobox-v2 placeholder="Select Item" i18n-placeholder
+                 [overlayConfig]="{ minHeight: 310 }"
```
Blakesalot, line 2: This doesn't seem quite right to have consumers arbitrarily set the overlay config min height to this magic number. Do you mind trying to preserve the calculateContainerHeight functionality and make it work to somehow avoid this?

packages/bits/demo/src/components/demo/select-v2/select-v2-virtual-scroll/select-v2-virtual-scroll.example.component.html
```diff
@@ -1,4 +1,5 @@
 <nui-select-v2 placeholder="Select Item" i18n-placeholder
+               [overlayConfig]="{ minHeight: 310 }"
```
Blakesalot, line 2: Same comment as above.

packages/bits/demo/src/components/demo/combobox-v2/combobox-v2-virtual-scroll/combobox-v2-virtual-scroll.example.component.html
DmytroPutrin: add getter and comment for this, thanks.

Blakesalot:
@DmytroPutrin, is there a way we could automatically handle this in the select-v2 and combobox-v2 components? It would be nice if the consumer didn't have to worry about it at all.

packages/bits/demo/src/components/demo/combobox-v2/combobox-v2-virtual-scroll/combobox-v2-virtual-scroll.example.component.html
DmytroPutrin: @Blakesalot I'm not sure that root component (Select in our case) should look inside content projection. But, I recognize that my solution still doesn't work exactly right, so I provide another solution for it. Please let me know if new approach applicable for us.

packages/bits/src/lib/select-v2/base-select-v2.ts
```diff
@@ -433,4 +444,29 @@ export abstract class BaseSelectV2 implements AfterViewInit, AfterContentInit, C
             resizeObserver.unobserve(this.elRef.nativeElement);
         });
     }
+
+    /** This helps to dynamically set minHeight for overlay to avoid issues with double scroll
+     Overlay minHeight should be bigger than cdkVirtualScroll container
+     */
+    private detectVirtualScroll(): void {
```
Blakesalot, line 451: Maybe a more self-documenting name like `adjustDropdownOnVScrollResize`?

packages/bits/src/lib/select-v2/base-select-v2.ts
```diff
@@ -433,4 +444,29 @@ export abstract class BaseSelectV2 implements AfterViewInit, AfterContentInit, C
             resizeObserver.unobserve(this.elRef.nativeElement);
         });
     }
+
+    /** This helps to dynamically set minHeight for overlay to avoid issues with double scroll
+     Overlay minHeight should be bigger than cdkVirtualScroll container
+     */
```
Blakesalot, line 450: JSDoc formatting adjustment:
```suggestion
    /**
     * This helps to dynamically set minHeight for overlay to avoid issues with double scroll
     * Overlay minHeight should be bigger than cdkVirtualScroll container
     */
```

packages/bits/src/lib/select-v2/base-select-v2.ts
```diff
@@ -138,6 +143,7 @@ export abstract class BaseSelectV2 implements AfterViewInit, AfterContentInit, C
     private _selectedOptions: SelectV2OptionComponent[] = [];
 
     private _ariaLabel: string = "";
+    private _virtualScrollResizeObserver: ResizeObserver;
```
Blakesalot, line 146: By convention, Nova typically uses leading underscores only for private members with a corresponding public getter/setter.

```suggestion
    private virtualScrollResizeObserver: ResizeObserver;
```

packages/bits/src/lib/select-v2/base-select-v2.ts
DmytroPutrin: renamed

packages/bits/src/lib/select-v2/base-select-v2.ts
DmytroPutrin: renamed it, thanks

packages/bits/demo/src/components/demo/combobox-v2/combobox-v2-virtual-scroll/combobox-v2-virtual-scroll.example.component.html
```diff
@@ -1,6 +1,6 @@
 <nui-combobox-v2 placeholder="Select Item" i18n-placeholder
                  [formControl]="comboboxControl"
-                 [isTypeaheadEnabled]="false"
```
Blakesalot, line 3: Is this change necessary?

packages/bits/demo/src/components/demo/combobox-v2/combobox-v2-virtual-scroll/combobox-v2-virtual-scroll.example.component.ts
```diff
@@ -39,23 +38,12 @@ export class ComboboxV2VirtualScrollExampleComponent implements OnDestroy, After
         });
 
         this.combobox.valueChanged.pipe(
-            tap(v => this.filteredItems = of(this.filterItems(v as string))),
-            // should recalculate container height after filtered items will be rendered
             delay(0),
-            tap(this.calculateContainerHeight),
+            tap(v => this.calculateContainerHeight()),
```
Blakesalot, line 42: Is this change necessary? If so, I'm curious why.

packages/bits/src/lib/select-v2/base-select-v2.ts
```diff
@@ -433,4 +444,30 @@ export abstract class BaseSelectV2 implements AfterViewInit, AfterContentInit, C
...
+     * scroll. Overlay minHeight should be bigger than cdkVirtualScroll container.
+     */
+    private adjustDropdownOnVScrollResize(): void {
+        if (!this.cdkVirtualScroll) {
+            return;
+        }
+
+        const element = this.cdkVirtualScroll.elementRef.nativeElement;
+        const height = parseInt(element.style.height, 10);
+        const minHeight = Number.isNaN(height) ? 0 : height + 10;
```
Blakesalot, line 459: Can we use a named constant for the 10 to make it more self-documenting? Maybe `vScrollHeightBuffer`? This constant can also be used at line 465.

packages/bits/src/lib/select-v2/base-select-v2.ts
Blakesalot: Added it [here](https://github.com/solarwinds/nova/pull/276/files#diff-d2adc1eebb3bcd5bf6298ad22b60cccfee0e21315462af82bbdd40aa4b073f57)

packages/bits/demo/src/components/demo/combobox-v2/combobox-v2-virtual-scroll/combobox-v2-virtual-scroll.example.component.ts
```diff
@@ -39,23 +37,12 @@ export class ComboboxV2VirtualScrollExampleComponent implements OnDestroy, After
         });
 
         this.combobox.valueChanged.pipe(
-            tap(v => this.filteredItems = of(this.filterItems(v as string))),
```
Blakesalot, line 42: @DmytroPutrin, I'm not sure, but I think this is needed to filter out items that don't match the string currently entered into the combobox.

Blakesalot: But, maybe this is what you said Dima G. found a bug about, and this is part of the fix?

Blakesalot: As long as typeahead successfully filters out the non-matching strings, I think we're good.

ivankozlovskyi (approved):

PR #280, opened 2021-05-12
Bugfix/ NUI-6037 fixed 'ObjectUnsubscribedErrorImpl' on destroy Wizard hook

dmytrognatyshyn:


DmytroPutrin (dismissed):

packages/bits/src/lib/wizard-v2/wizard-step.component.ts
```diff
@@ -126,4 +126,9 @@ export class WizardDirective extends CdkStepper implements AfterContentInit {
             }
         });
     }
+
+    ngOnDestroy() {
+        this._destroyed.next();
+        this._destroyed.complete();
```
Blakesalot, line 132: We also need to call super.ngOnDestroy() here.

Here's this for reference: https://github.com/solarwinds/nova/blob/703da1a016dc3bf307d1898a59c3f35076150476/docs/STYLE_GUIDE.md#ngondestroy-and-component-inheritance

PR #283, opened 2021-05-13
NUI-4907-Tests-Review-and-Improvements-for-FormField-component

elenates:
- Deleted a few e2e tests which are duplicated by visual tests.
- Disabled "should change model for dateTimePicker" test because of bug (should be reviewed after NUI-6067 fixed).

### DIFF
--- a/packages/bits/spec/components/form-field/form-field.e2e.ts
+++ b/packages/bits/spec/components/form-field/form-field.e2e.ts
@@ -1,7 +1,7 @@
-import {browser, by, element, ElementFinder } from "protractor";
+import { browser, by, element, ElementFinder } from "protractor";
 
-import {Atom} from "../../atom";
-import {Helpers} from "../../helpers";
+import { Atom } from "../../atom";
+import { Helpers } from "../../helpers";
 import {
     ButtonAtom,
     CheckboxAtom,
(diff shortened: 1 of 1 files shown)


packages/bits/spec/components/form-field/form-field.e2e.ts
```diff
@@ -227,21 +196,20 @@ describe("USERCONTROL form-field >", () => {
             await timepicker.icon.getElement().click();
             expect(await timepicker.textbox.input.getId())
                 .toEqual(await (await browser.switchTo().activeElement()).getId());
-
             await timepicker.icon.getElement().click();
         });
 
-        it("should change model for dateTimePicker", async () => {
```
Blakesalot, line 234: Does this old test work as it was previously written? If so, it might be good to restore and enable the old version and create a new disabled test that recreates the bug you found.

elenates: Blake, no, the test was false negative, so it doesn't work correctly and was shown as passed every time, though the error was there. That's why we didn't found the bug before.
So I would like to rewrite the test to cover both situations.

Blakesalot (approved):

baggiosw (approved):

madelyneh (approved):

PR #284, opened 2021-05-13
Feature/n UI 6013

DmytroPutrin:


packages/bits/src/lib/select-v2/combobox-v2/combobox-v2.component.less
```diff
@@ -71,6 +71,11 @@
             border-right: 1px solid;
             .setCssVariable(border-right-color, nui-color-line-default);
         }
+
+        &:focus {
+            border-radius: 6px;
```
Blakesalot, line 76: 6px seems like a strange number perhaps. Was 6px suggested by UX? Also, a file-scoped variable for this at the top of the file might be good.

packages/bits/src/lib/select-v2/constants.ts
```diff
@@ -4,3 +4,5 @@ import { IOptionedComponent } from "./types";
 
 // Will be renamed in scope of the NUI-5797
 export const NUI_SELECT_V2_OPTION_PARENT_COMPONENT = new InjectionToken<IOptionedComponent>("NUI_SELECT_V2_OPTION_PARENT_COMPONENT");
+export const ANNOUNCER_OPEN_MESSAGE = (optionsLength: number) => `${optionsLength} options available`;
+export const ANNOUNCER_CLOSE_MESSAGE = "Dropdown closed";
```
Blakesalot, line 8: @DmytroPutrin, can you check whether the LivaAnnouncer supports i18n? If so, it would be good to go ahead and implement that or create a ticket to do that. Even if LiveAnnouncer doesn't support it, I think it would be good to have a ticket for researching how to internationalize all of our a11y support.

packages/bits/src/lib/select-v2/constants.ts
```diff
@@ -4,3 +4,5 @@ import { IOptionedComponent } from "./types";
 
 // Will be renamed in scope of the NUI-5797
 export const NUI_SELECT_V2_OPTION_PARENT_COMPONENT = new InjectionToken<IOptionedComponent>("NUI_SELECT_V2_OPTION_PARENT_COMPONENT");
+export const ANNOUNCER_OPEN_MESSAGE = (optionsLength: number) => `${optionsLength} options available`;
```
Blakesalot, line 7: It would be better to have this as just a vanilla constant. We don't normally use the constant naming convention (`CAPS_WITH_UNDERSCORES`) for functions in Nova. How about the following?

```suggestion
export const ANNOUNCER_OPEN_MESSAGE_SUFFIX = "options available";
```

Then you can use the template literal syntax directly where ever this suffix is needed.

packages/bits/src/lib/select-v2/option-key-control.service.ts
```diff
@@ -111,4 +117,18 @@ export class OptionKeyControlService<T extends IOption> {
...
+    private announceNavigatedOption(): void {
+        const activeItem = this.keyboardEventsManager.activeItem;
+
+        if (activeItem) {
+            this.liveAnnouncer.announce((activeItem as any).value);
+        }
+    }
+
+    private announceDropdown(open: boolean) {
+        const message = open ? ANNOUNCER_OPEN_MESSAGE(this.optionItems.length) : ANNOUNCER_CLOSE_MESSAGE;
```
Blakesalot, line 130: See my previous comment for context:
```suggestion
        const message = open ? `${this.optionItems.length} ${ANNOUNCER_OPEN_MESSAGE_SUFFIX}` : ANNOUNCER_CLOSE_MESSAGE;
```

packages/bits/src/lib/select-v2/selected-items-key-control.service.ts
```diff
@@ -1,4 +1,4 @@
-import { ActiveDescendantKeyManager } from "@angular/cdk/a11y";
+import { ActiveDescendantKeyManager, LiveAnnouncer } from "@angular/cdk/a11y";
```
Blakesalot, line 1: It might be good to add some unit tests for the changes in this file.

packages/bits/src/lib/select-v2/option-key-control.service.ts
```diff
@@ -1,9 +1,10 @@
-import { ActiveDescendantKeyManager } from "@angular/cdk/a11y";
+import { ActiveDescendantKeyManager, LiveAnnouncer } from "@angular/cdk/a11y";
```
Blakesalot, line 1: It might be good to add some unit tests for the changes in this file.

packages/bits/src/lib/select-v2/combobox-v2/combobox-v2.component.less
DmytroPutrin: We don't have spec for this focus, because of overflow styles inside combobox. I used mixin with inner shadow and make. I set 6px for radius to avoid rectangular focus. But I'll clarify it with ux.

packages/bits/src/lib/select-v2/constants.ts
DmytroPutrin: As I could see no - cdk insert element which screen reader reads inside dom without internalization. I created ticked (NUI-2139) for this.

packages/bits/src/lib/select-v2/constants.ts
DmytroPutrin: resolved it, thanks

packages/bits/src/lib/select-v2/selected-items-key-control.service.ts
DmytroPutrin: add spec file for this service.

packages/bits/src/lib/select-v2/option-key-control.service.ts
DmytroPutrin: we will get rid of keyCode property across all project (NUI-2492). I guess ut for this file better create in scope of this ticket avoid additional work.

packages/bits/src/lib/select-v2/combobox-v2/combobox-v2.component.less
DmytroPutrin: clarify with ux and reduce border radius to 3px

ivankozlovskyi (dismissed):

dmytrognatyshyn (approved):

ivankozlovskyi (approved):

dmytrognatyshyn (approved):

PR #300, opened 2021-05-24
moving language set to the component

dmytrofirsov:


ivankozlovskyi (approved):

DmytroPutrin (approved):

Vita-Karas (approved):

Blakesalot (approved):

packages/bits/src/lib/docs/example-wrapper/example-wrapper.component.ts
```diff
@@ -47,7 +49,13 @@ export class ExampleWrapperComponent implements OnInit {
         private sourcesService: SourcesService,
         @Inject(DOCUMENT) private document: Document,
         private plunkerProjectService: PlunkerProjectService
-    ) {}
+    ) {
+        hljs.registerLanguage("typescript", require("highlight.js/lib/languages/typescript"));
```
Blakesalot, line 53: @dmytrofirsov, is it okay to call `registerLanguage` every time a new example wrapper is instantiated? I'm assuming subsequent calls are a no-op, but I wanted to make sure.

dmytrofirsov: @Blakesalot I was hesitating over this tbh. Hljs team does not check whether the language has already been registered but since they do not seem to do any heavy lifting there I decided to go this way, as it minimizes efforts for the consumers, and example wrappers are mostly used for the documentation. If for some reason the performance will ever become a problem here, we can always revert this to registering languages once on app level, but this will force users to not forget doing the same thing on their side.

Blakesalot: Okay, thanks for the update, Dima. Works for me :)

PR #305, opened 2021-05-25
NUI-4910 Tests Review and Improvements for menu component

elenates:
- deleted several e2e tests which were duplicated by visual tests,
- added a component to visual test to reduce time of test running,
- renamed one section to use it in visual test instead of e2e test.

### DIFF
--- a/packages/bits/demo/src/components/demo/menu/menu-visual-test/menu-visual-test.component.ts
+++ b/packages/bits/demo/src/components/demo/menu/menu-visual-test/menu-visual-test.component.ts
@@ -16,7 +16,7 @@ export class MenuVisualTestComponent {
                 { title: "Item 3", itemType: "action", action: this.actionWithParams.bind(this, 2) },
                 { title: "Item 4", itemType: "action", action: this.actionDone },
                 { title: "Item 5", itemType: "action", action: this.actionDone },
-                { title: "Item 6", itemType: "action", action: this.actionDone },]
+                { title: "Item 6", itemType: "action", action: this.actionDone },],
         },
         {
             itemsSource: [
@@ -27,7 +27,7 @@ export class MenuVisualTestComponent {
                 { title: "Item 11", itemType: "action", action: this.actionDone },
                 { title: "Item 12", itemType: "action", action: this.actionDone },
                 { title: "Item 13", itemType: "action", action: this.actionDone, disabled: true },
-                { title: "Item 14", itemType: "action", action: this.actionDone, itemClass: MenuActionType.destructive },]
+                { title: "Item 14", itemType: "action", action: this.actionDone, itemClass: MenuActionType.destructive },],
         },
     ];
 
@@ -43,7 +43,7 @@ export class MenuVisualTestComponent {
                 { title: "Menu item with checkbox", itemType: "option", disabled: true },
                 { title: "Menu item with icon", itemType: "action", icon: "table" },
                 { title: "Link menu item", itemType: "link", url: "#button", disabled: true },
-                { title: "Export PDF", itemType: "link", icon: "export-pdf", url: "#button" },]
+                { title: "Export PDF", itemType: "link", icon: "export-pdf", url: "#button" },],
         },
         {
             header: "section 2 title",
--- a/packages/bits/demo/src/components/demo/menu/menu-visual-test/menu-visual-test.component.ts
+++ b/packages/bits/demo/src/components/demo/menu/menu-visual-test/menu-visual-test.component.ts
@@ -40,6 +40,13 @@ export class MenuVisualTestComponent {
             {title: "Link menu item", itemType: "link", url: "#button", disabled: true},
             {title: "Export PDF", itemType: "link", icon: "export-pdf", url: "#button"},
         ]},
+        {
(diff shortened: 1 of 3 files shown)


packages/bits/demo/src/components/demo/menu/menu-visual-test/menu-visual-test.component.ts
```diff
@@ -40,6 +40,13 @@ export class MenuVisualTestComponent {
             {title: "Link menu item", itemType: "link", url: "#button", disabled: true},
             {title: "Export PDF", itemType: "link", icon: "export-pdf", url: "#button"},
         ]},
+        {
+            header: "section 2 title", itemsSource: [
```
Blakesalot, line 44: Newline preferred here. Also, do you mind updating the format of the other objects as well at lines 11, 19, and 32?

```suggestion
            header: "section 2 title",
            itemsSource: [
```

Blakesalot (dismissed):

Blakesalot (dismissed):

madelyneh (dismissed):

Blakesalot (approved):

madelyneh (approved):

packages/bits/spec/components/menu/menu.e2e.ts
```diff
@@ -30,82 +23,7 @@ describe("USERCONTROL Menu", () => {
...
-            const disabledOptionItem = menu.getMenuItemByContainingText("Disabled option");
-            const checkbox: CheckboxAtom = Atom.findIn(CheckboxAtom, disabledOptionItem.getElement());
-            expect(await checkbox.isChecked()).toBe(false);
-        });
-        it("should locate disabled switch", async () => {
-            const disabledSwitchItem = menu.getMenuItemByContainingText("Disabled switch");
-            const switchAtom = Atom.findIn(SwitchAtom, disabledSwitchItem.getElement());
-            expect(await switchAtom.isOn()).toBe(false);
-        });
-    });
```
ivankozlovskyi, line 91: It seems like there are no such cases in visual tests

packages/bits/spec/components/menu/menu.e2e.ts
```diff
@@ -130,22 +47,7 @@ describe("USERCONTROL Menu", () => {
...
 
     describe("> key navigation", () => {
-        const checkElementsActiveState = async () => {
-            await menu.toggleMenu();
-            await browser.actions().sendKeys(protractor.Key.ARROW_DOWN).perform();
-            expect(await menu.getMenuItemByIndex(0).isActiveItem()).toBe(true);
-            await browser.actions().sendKeys(protractor.Key.ARROW_DOWN).perform();
-            expect(await menu.getMenuItemByIndex(1).isActiveItem()).toBe(true);
-            await browser.actions().sendKeys(protractor.Key.ARROW_DOWN).perform();
-            expect(await menu.getMenuItemByIndex(2).isActiveItem()).toBe(true);
```
ivankozlovskyi, line 140: I think that only last one should be removed

packages/bits/spec/components/menu/menu.e2e.ts
```diff
@@ -202,14 +107,12 @@ describe("USERCONTROL Menu", () => {
                 });
 
                 it("should select and close menu when selecting menu action item", async () => {
-                    await menu.toggleMenu();
```
ivankozlovskyi, line 205: have you checked that menu is opened in this case, it is possible that those tests where  "await menu.toggleMenu();" was deleted become false positive

elenates: Discussed offline - "await menu.toggleMenu();" was moved to _beforeEach_

packages/bits/spec/components/menu/menu.e2e.ts
elenates: Discussed offline - _Disabled switch_, _Disabled action_, _Disabled link_ and _Disabled option_ are verified in another visual tests.

packages/bits/spec/components/menu/menu.e2e.ts
Blakesalot: I think Ivan may have meant that the test itself shouldn't be deleted, but we should keep it and just delete lines 139 and 140. Do I have that right, Ivan? But, maybe there is a visual test that covers the whole scenario?

Blakesalot:
@elenates, I just noticed that the browser gets refreshed with every test. I wonder if it would be an easy task to put the preparBrowser in a beforeAll instead of a beforeEach. Each test might have to clean up after itself, e.g. restore a menu to its default state somehow, or maybe there could be some generic cleanup in an afterEach. If it's too much trouble, no worries. I was just thinking of a way to speed the test suite up by avoiding unnecessary page refreshes.

packages/bits/spec/components/menu/menu.e2e.ts
elenates: Hi Blake, sorry, we discussed it offline with Ivan, and I didn't write it. Yes, the visual test covers the whole scenario anyway.

elenates:
@Blakesalot, thank you for your comment, that a great idea. I am working on that, and looks like we can speed the test up from 8 to 3 secs.

PR #309, opened 2021-05-26
Feature/n UI 1566 restore wizard state

DmytroPutrin:
- add restore wizard state functional
- update wizard-v2 docs page

### DIFF
--- a/packages/bits/demo/src/components/demo/wizard-v2/wizard-restore-state/wizard-restore-state.example.component.html
+++ b/packages/bits/demo/src/components/demo/wizard-v2/wizard-restore-state/wizard-restore-state.example.component.html
@@ -175,7 +175,7 @@
 </ng-template>
 
 <ng-template #dynamicTemplate1>
-    <p i18n>Hi! I was created dynamically, because you've checked the checkbox on the previous step! Check it off to hide me!</p>
+    <p i18n>Hi! I was created dynamically, because you've checked the checkbox on the previous step! Uncheck it to hide me!</p>
 </ng-template>
 
 <ng-template #dynamicTemplate2>
--- a/packages/bits/demo/src/components/demo/wizard-v2/wizard-restore-state/wizard-restore-state.example.component.html
+++ b/packages/bits/demo/src/components/demo/wizard-v2/wizard-restore-state/wizard-restore-state.example.component.html
@@ -25,14 +25,16 @@
                     </nui-validation-message>
                 </nui-form-field>
 
-                <nui-checkbox formControlName="privacy" required>
-                    Privacy Policy
-                </nui-checkbox>
+                <nui-form-field [control]="form.get(['personDetails','privacy'])">
+                    <nui-checkbox formControlName="privacy">
+                        Privacy Policy
+                    </nui-checkbox>
+                </nui-form-field>
 
-                    <ng-template wizardStepFooter>
-                        <ng-container *ngTemplateOutlet="wizardCancel"></ng-container>
-                        <ng-container *ngTemplateOutlet="wizardNext"></ng-container>
-                    </ng-template>
+                <ng-template wizardStepFooter>
+                    <ng-container *ngTemplateOutlet="wizardCancel"></ng-container>
+                    <ng-container *ngTemplateOutlet="wizardNext"></ng-container>
+                </ng-template>
             </nui-wizard-step-v2>
 
             <nui-wizard-step-v2 label="Organization"
--- a/packages/bits/demo/src/components/demo/wizard-v2/wizard-restore-state/wizard-restore-state.example.component.html
+++ b/packages/bits/demo/src/components/demo/wizard-v2/wizard-restore-state/wizard-restore-state.example.component.html
@@ -27,7 +27,7 @@
                 </nui-form-field>
 
                 <nui-form-field [control]="form.get(['personDetails','privacy'])">
-                    <nui-checkbox formControlName="privacy">
+                    <nui-checkbox formControlName="privacy" [value]="true">
                         Privacy Policy
                     </nui-checkbox>
                 </nui-form-field>
--- a/packages/bits/demo/src/components/demo/wizard-v2/wizard-restore-state/wizard-restore-state.example.component.html
+++ b/packages/bits/demo/src/components/demo/wizard-v2/wizard-restore-state/wizard-restore-state.example.component.html
@@ -48,7 +48,7 @@
                                 formGroupName="organization"
                                 [stepControl]="form.get(['organization'])">
                 <nui-form-field [control]="form.get(['organization', 'title'])"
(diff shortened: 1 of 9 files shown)


packages/bits/demo/src/components/demo/wizard-v2/wizard-restore-state/wizard-restore-state.example.component.html
```diff
@@ -23,12 +23,16 @@
                                             i18n>
                         This is required
                     </nui-validation-message>
+                </nui-form-field>
+
+                <nui-checkbox formControlName="privacy" required>
```
Blakesalot, line 28: it seems like this should be inside a `nui-form-field` as well?

packages/bits/demo/src/components/demo/wizard-v2/wizard-restore-state/wizard-restore-state.example.component.html
```diff
@@ -23,12 +23,16 @@
...
+                    Privacy Policy
+                </nui-checkbox>
 
                     <ng-template wizardStepFooter>
                         <ng-container *ngTemplateOutlet="wizardCancel"></ng-container>
                         <ng-container *ngTemplateOutlet="wizardNext"></ng-container>
                     </ng-template>
```
Blakesalot: indentation adjustment needed

packages/bits/demo/src/components/demo/wizard-v2/wizard-restore-state/wizard-restore-state.example.component.html
DmytroPutrin: done

packages/bits/demo/src/components/demo/wizard-v2/wizard-restore-state/wizard-restore-state.example.component.html
DmytroPutrin: fixed, thanks

ivankozlovskyi (approved):

Vita-Karas (approved):

packages/bits/src/lib/wizard-v2/wizard.directive.ts
```diff
@@ -1,5 +1,6 @@
 import {
```
Blakesalot, line 1: @DmytroPutrin, do you mind adding some unit tests for the new functionality?

packages/bits/demo/src/components/demo/wizard-v2/wizard-restore-state/wizard-restore-state.example.component.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    public openDialog(content: TemplateRef<string>): void {
+        this.activeDialog = this.dialogService.open(content, {size: "lg"});
+    }
+
+    public closeDialog(): void {
+        this.activeDialog.close();
+    }
+
+    public onWizardDestroy(steps: QueryList<WizardStepV2Component>): void {
```
lukasholcik, line 45: Why is this callback called `onWizardDestroy`?

packages/bits/src/lib/wizard-v2/wizard.directive.ts
DmytroPutrin: in case we have no spec files for wizard 2, here separate task for ut (NUI-5070).

packages/bits/demo/src/components/demo/wizard-v2/wizard-restore-state/wizard-restore-state.example.component.ts
DmytroPutrin: curranty, we emitting steps list (which we can restore later)  onDestroy hook because it most common case for users.

packages/bits/src/lib/wizard-v2/public-api.ts
```diff
@@ -10,3 +13,7 @@ export * from "./wizard-horizontal/wizard-horizontal.component";
 export * from "./wizard-step/wizard-step.component";
 export * from "./wizard-step-header/wizard-step-header.component";
 export * from "./wizard-vertical/wizard-vertical.component";
+
+export interface IWizardState {
```
lukasholcik, line 17: Please create a dedicated types.ts file with this interface. public-api.ts should not directly contain any type / class definitions.

packages/bits/demo/src/components/demo/wizard-v2/wizard-restore-state/wizard-restore-state.example.component.html
```diff
@@ -0,0 +1,187 @@
...
+                <button type="button" nui-button (click)="closeDialog()">
+                    Cancel
+                </button>
+            </ng-template>
+        </nui-wizard-horizontal>
+    </form>
+</ng-template>
+
+<ng-template #dynamicTemplate1>
+    <p i18n>Hi! I was created dynamically, because you've checked the checkbox on the previous step! Check it off to hide me!</p>
```
Blakesalot, line 178:
```suggestion
    <p i18n>Hi! I was created dynamically, because you've checked the checkbox on the previous step! Uncheck it to hide me!</p>
```

dmytrofirsov:
@Blakesalot  @lukasholcik please check whether this solution is better. It now stores the state only if the wizard is finished, if not user has to go step by step, just like JR requested. I adjusted the example as well to show that the dynamically added steps are also preserved on wizard restore since it all is being stored in forms.

packages/bits/demo/src/components/demo/wizard-v2/wizard-restore-state/wizard-restore-state.example.component.html
```diff
@@ -0,0 +1,187 @@
...
+                                caption="Organization">
+                    <nui-textbox formControlName="title"></nui-textbox>
+                    <nui-validation-message for="required"
+                                            *ngIf="form.get(['organization','title'])?.errors?.required"
+                                            i18n>
+                        This is required
+                    </nui-validation-message>
+                </nui-form-field>
+                <nui-form-field caption="Date"
+                                class="d-block my-4"
```
Blakesalot, line 62: The form fields in the old wizard examples just used `mb-4`, so it seems like this might add too much space between the form fields when they're right next to each other. Here's a form field in the old wizard docs: https://github.com/solarwinds/nova/blob/98ef229973f8b1903efa77c902b2cc2c2f5cc9b1/packages/bits/demo/src/components/demo/wizard/wizard-reset-step/wizard-reset-step.example.component.html#L33

dmytrofirsov: fixed

Vita-Karas (approved):

ivankozlovskyi (approved):

PR #310, opened 2021-05-26
bugfix/NUI-6061-Focused-checkbox-overlaps-header

Vita-Karas:


DmytroPutrin (dismissed):

packages/bits/spec/components/table/table.visual.ts
```diff
@@ -80,6 +83,12 @@ describe(`Visual tests: ${name}`, () => {
         await camera.say.cheese("Edit columns");
         await expanders.customActions.click();
 
+        await expanders.selectPinnedHeader.click();
+        await selectPinnedHeaderTable.getCell(1, 0).click();
+        await browser.executeScript("document.getElementById('table-select-pinned-header').getElementsByClassName('nui-table__container')[0].scrollTop = '20'");
```
dmytrofirsov, line 88: we may want to have this in a bit generic format in our Helpers, so we could reuse it. Something like scrollUp() or scrollDown()

Vita-Karas: I think it is not a generic script, which we can reuse in our Helpers

ivankozlovskyi (approved):

packages/bits/src/styles/mixins/focus.less
```diff
@@ -1,6 +1,5 @@
 .focus-outline-base() {
     outline: none;
-    z-index: @zindex-active;
```
Blakesalot, line 3: @Vita-Karas, I would leave this here since it's been there for a couple of years. It seems too risky to remove it. It might be good to write a ticket to consider removing it on a major Nova version bump, though.

packages/bits/src/lib/checkbox/checkbox.component.less
```diff
@@ -50,7 +50,8 @@
         }
 
         &:focus {
-            .focus-outline-inset();
+            outline: none;
```
Blakesalot, line 53: I loaded this up into the browser and it looks like chrome doesn't like this value for outline. As a result when you focus one checkbox with the mouse and then hit TAB, the black outline shows up. The fix is to change `none` to `0`.

```suggestion
            outline: 0;
```

packages/bits/src/styles/mixins/focus.less
Vita-Karas: Created ticket: [NUI-6103](https://jira.solarwinds.com/browse/NUI-6103)

ivankozlovskyi (dismissed):

packages/bits/src/lib/checkbox/checkbox.component.less
```diff
@@ -50,7 +50,8 @@
         }
 
         &:focus {
-            .focus-outline-inset();
+            outline: 0;
+            box-shadow: inset 0 0 0 @nui-line-medium fade(@nui-color-selected-contrast, 50%);
```
dmytrofirsov, line 54: please add a comment here to switch back to mixin once that z-index is removed

packages/bits/src/styles/mixins/focus.less
```diff
@@ -1,5 +1,6 @@
 .focus-outline-base() {
     outline: none;
+    // TODO: will removed in scope of NUI-6103
```
dmytrofirsov, line 3: "will **be** removed"

Vita-Karas, line 3:
```suggestion
    // TODO: will be removed in scope of NUI-6103
```

packages/bits/src/lib/checkbox/checkbox.component.less
```diff
@@ -50,7 +50,8 @@
         }
 
         &:focus {
-            .focus-outline-inset();
+            outline: 0;
```
Vita-Karas, line 53:
```suggestion
            // TODO: switch back to mixin .focus-outline-inset() when z-index is removed in the scope of NUI-6103
            outline: 0;            
```

dmytrofirsov (approved):

ivankozlovskyi (approved):

packages/bits/src/lib/checkbox/checkbox.component.less
dmytrofirsov: I'm pretty sure `none` is a valid value for the `outline`.
https://www.w3schools.com/css/css_outline.asp

image
image

Blakesalot: Okay, thanks for checking. I'm not sure what I was seeing then.

PR #311, opened 2021-05-27
Feature/n UI 5066 allow header icons customization in header

dmytrognatyshyn:


packages/bits/demo/src/components/demo/wizard-v2/wizard-custom-layout/wizard-custom.component.less
```diff
@@ -10,11 +10,16 @@
...
-
         .setCssVariable(border-color, nui-color-line-default);
         .setCssVariable(background, nui-color-bg-content);
-
+        flex-direction: row-reverse;
+        justify-content: flex-end;
         padding: 0 8px;
+
+        .nui-icon{
+            transform: translate(-70%, -55%);
```
Blakesalot, line 20: The consumer really shouldn't have to reposition an icon like this. I think a further investigation on the root cause of the misalignment would be better. It may have something to do with the fact that the circle used for the icon is 20px instead of the typical icon size of 16px.

packages/bits/demo/src/components/demo/wizard-v2/wizard-custom-layout/wizard-custom.component.less
```diff
@@ -23,7 +28,11 @@
         color: @nui-color-text-light;
         border-color: transparent;
 
-        svg path {
+        .nui-wizard-step-label{
+            color: @nui-color-text-inverse;
+        }
+
+        .nui-step-icon-selected svg circle, .nui-step-icon-selected svg path {
             fill: @nui-color-text-inverse;
```
Blakesalot, line 36: These color's aren't dark mode friendly. To make them dark mode friendly, we should use the setCssVariable mixin.

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.less
```diff
@@ -62,6 +88,7 @@
     text-overflow: ellipsis;
     min-width: @wizard-label-min-width;
     vertical-align: middle;
+    color: @nui-color-text-secondary;
```
Blakesalot, line 91: needs setCssVariable mixin to be theme-friendly

packages/bits/src/lib/wizard-v2/wizard-icon.directive.ts
```diff
@@ -0,0 +1,26 @@
...
+    index: number;
+    active: boolean;
+    optional: boolean;
+}
+
+export interface MatStepperIconContext {
+    index: number;
+    active: boolean;
+    optional: boolean;
+}
```
Blakesalot, line 15: Do we need both of these since they're identical?

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.html
```diff
@@ -1,19 +1,8 @@
...
-    </div>
-</div>
-<div class="nui-step-label"
-     [class.nui-step-label-active]="active"
-     [class.nui-step-label-selected]="selected"
-     [class.nui-step-label-error]="state == 'error'">
+<div class="nui-wizard-step-label"
+     [class.nui-wizard-step-label-active]="completed"
+     [class.nui-wizard-step-label-selected]="selected"
+     [class.nui-wizard-step-label-error]="state == 'error'">
```
Blakesalot, line 5: we should use BEM here. e.g.

```suggestion
     [class.nui-wizard-step-header__step-label--active]="completed"
     [class.nui-wizard-step-header__step-label--selected]="selected"
     [class.nui-wizard-step-header__step-label--error]="state == 'error'">
```

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.ts
```diff
@@ -13,9 +21,9 @@ import {WizardStepLabelDirective} from "../wizard-step-label.directive";
     host: {
         "class": "nui-wizard-step-header mat-focus-indicator",
         "[class.nui-wizard-step-header-selected]": "selected",
-        "[class.nui-wizard-step-header-active]": "active",
         "[class.nui-wizard-step-header-optional]": "optional",
         "[class.nui-wizard-step-header-completed]": "state == 'done'",
+        "[class.nui-wizard-step-header-disabled]": "!completed",
```
Blakesalot, line 26: We should use BEM here (see other comment)

packages/bits/src/lib/wizard-v2/wizard-vars.less
```diff
@@ -3,14 +3,15 @@
...
 
-@wizard-side-gap: 24px;
-@wizard-line-width: 1px;
+@wizard-side-gap: 15px;
+@wizard-line-width: 2px;
 @wizard-line-gap: 0px;
 
 @wizard-step-sub-label-font-size: 12px;
-@wizard-step-header-icon-size: 12px;
+@wizard-step-header-icon-size: 16px;
```
Blakesalot, line 17: nui-framework-icon-sizes.less already has a variable for this: @icon-size-default

packages/bits/src/lib/wizard-v2/wizard-vars.less
```diff
@@ -3,14 +3,15 @@
...
 @wizard-label-header-height: 24px;
+@wizard-label-header-width: 145px;
 @wizard-label-position-bottom-top-gap: 16px;
 @wizard-label-min-width: 50px;
 
 @wizard-vertical-content-margin: 36px;
 
-@wizard-side-gap: 24px;
-@wizard-line-width: 1px;
+@wizard-side-gap: 15px;
```
Blakesalot, line 12: nui-framework-spaces.less already has a variable for this value: @nui-space-md

packages/bits/src/lib/wizard-v2/wizard-vars.less
```diff
@@ -3,14 +3,15 @@
...
+@wizard-line-width: 2px;
```
Blakesalot, line 13: nui-framework-line-thickness.less has a variable for this one already too: @nui-line-medium

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.less
```diff
@@ -33,12 +39,32 @@
...
+        .setCssVariable(fill, nui-color-text-secondary);    }
+
+
+    &.nui-step-icon-selected {
+        svg circle, svg path{
+            .setCssVariable(fill, nui-color-text-default);
+        }
+    }
+
+    &.nui-step-icon-state-done {
```
Blakesalot, line 62: we should use BEM for these classes as well. e.g.

```
&.nui-wizard-step-header__icon--selected {
```

packages/bits/src/lib/wizard-v2/wizard.component.less
```diff
@@ -145,7 +145,7 @@
 
 .nui-wizard-horizontal-content-container {
     overflow: hidden;
-    padding: 0 @wizard-side-gap @wizard-side-gap @wizard-side-gap;
+    padding: @wizard-side-gap;
```
ewok-janitor, line 148: In the grooming we had on Thursday, we talked about not having padding in the content area.

Will this be addressed in a subsequent ticket?

packages/bits/src/lib/wizard-v2/wizard-icon.directive.ts
```diff
@@ -0,0 +1,26 @@
...
+
+@Directive({
+    selector: "ng-template[nuiWizardIcon]",
+})
+export class NuiWizardIconDirective {
```
Blakesalot, line 20: We typically only prefix module names and selectors with `Nui`

```suggestion
export class WizardIconDirective {
```

Blakesalot, line 20: I wonder if a better name might be `WizardStepStateDirective`

packages/bits/src/lib/wizard-v2/wizard-icon.directive.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    /** Name of the icon to be overridden. */
+    @Input() nuiWizardIcon: StepState;
```
Blakesalot, line 23: Maybe this instead?

```suggestion
    @Input() stepState: StepState;
```

packages/bits/src/lib/wizard-v2/wizard-icon.directive.ts
dmytrognatyshyn: No, we don't need both. I forgot to remove

packages/bits/src/lib/wizard-v2/wizard-icon.directive.ts
dmytrognatyshyn: @Blakesalot I think WizardIconDirective is a better name for it because this directive is used only for overriding icons.

packages/bits/src/lib/wizard-v2/wizard-icon.directive.ts
dmytrognatyshyn: @Blakesalot Here I think it is a good idea to rename because this input is a step state in which icon should be replaced

packages/bits/demo/src/components/demo/wizard-v2/wizard-custom-layout/wizard-custom.component.less
```diff
@@ -10,24 +10,32 @@
...
         border-color: transparent;
 
-        svg path {
-            fill: @nui-color-text-inverse;
+        .nui-wizard-step-label{
+            .setCssVariable(color, nui-color-text-inverse)
+        }
+
+        .nui-wizard-step-header__icon--selected svg circle, .nui-wizard-step-header__icon--selected svg path {
+            .setCssVariable(fill, nui-color-text-inverse)
```
Blakesalot, line 31: Aren't we going to try using the `nui-icon` inputs to set the color?

Blakesalot: I guess we're not exposing those inputs to the wizard consumer. In this case, maybe we should avoid customizing the color in the example since we're not doing the full "puppy picture programming" implementation.

packages/bits/demo/src/components/demo/wizard-v2/wizard-custom-layout/wizard-custom.component.html
```diff
@@ -16,7 +16,7 @@
                         [attr.aria-label]="step.ariaLabel || null"
                         [attr.aria-labelledby]="(!step.ariaLabel && step.ariaLabelledby) ? step.ariaLabelledby : null"
                         [index]="i"
-                        [state]="_getIndicatorType(i, step.state)"
+                        [state]="getStepState(i, step.state)"
```
Blakesalot, line 19: It's not your fault, @dmytrognatyshyn, but the function calls in this template seem pretty mysterious. It looks like getStepState is defined in the wizard directive. Usually, I would expect a method invoked in a template to be a member of this component instead of some other entity. fyi, @dmytrofirsov

packages/bits/demo/src/components/demo/wizard-v2/wizard-custom-layout/wizard-custom.component.less
```diff
@@ -10,24 +10,32 @@
...
         }
     }
 
+    .nui-wizard-step-icon{
+        height: auto;
```
Blakesalot, line 36: what is this needed for exactly?

packages/bits/demo/src/components/demo/wizard-v2/wizard-docs/wizard-docs.component.html
```diff
@@ -57,3 +59,12 @@ <h3>Async Form Validation</h3>
 <nui-example-wrapper filenamePrefix="wizard-async-form-validation" exampleTitle="Async Form Validation">
     <nui-wizard-async-form-validation-example></nui-wizard-async-form-validation-example>
 </nui-example-wrapper>
+
+<h2>Wizard with custom icons</h2>
+<p>
+    You are able to add custom icons to wizard.
+</p>
+<h3>Wizard with custom icons</h3>
```
Blakesalot, line 67: seems like maybe a redundant (extra) header?

packages/bits/demo/src/components/demo/wizard-v2/wizard-docs/wizard-docs.component.html
```diff
@@ -57,3 +59,12 @@ <h3>Async Form Validation</h3>
...
+<nui-example-wrapper filenamePrefix="wizard-horizontal-layout" exampleTitle="Horizontal Layout Wizard">
```
Blakesalot, line 68: filenamePrefix needs updating

Blakesalot, line 68: the exampleTitle input value also needs updating

packages/bits/demo/src/components/demo/wizard-v2/wizard-custom-layout/wizard-custom.component.html
Blakesalot: ah, I understand now. this is an extension of the NuiWizardDirective

packages/bits/demo/src/components/demo/wizard-v2/wizard-custom-layout/wizard-custom.component.less
Blakesalot: If we keep this color override in some form, I think it should be nui-color-text-light instead of nui-color-text-inverse. It would make more sense to me.

packages/bits/demo/src/components/demo/wizard-v2/wizard-custom-layout/wizard-custom.component.less
```diff
@@ -10,24 +10,32 @@
...
+    .nui-wizard-step-header--selected {
         .setCssVariable(background-color, nui-color-active);
 
         color: @nui-color-text-light;
         border-color: transparent;
 
-        svg path {
-            fill: @nui-color-text-inverse;
+        .nui-wizard-step-label{
+            .setCssVariable(color, nui-color-text-inverse)
```
Blakesalot, line 27: If we keep the color overrides, I think we should use nui-color-text-light

Blakesalot: That said, maybe we can explore a more explicit way to set the colors?

dmytrofirsov: Agree, that 'inverse' color looks strange. I also agree with finding a more explicit way of setting the color of icons. Setting colors sounds more like a new task to me, which we decided to postpone on, didn't we? We may also need input from UX here since the specs lack details about colors. We requested wizard specs to be updated, so hopefully, we'll get some details soon.

packages/bits/demo/src/components/demo/wizard-v2/wizard-with-custom-icons/wizard-with-custom-icons-example.component.html
```diff
@@ -0,0 +1,40 @@
+<nui-wizard-horizontal [customIcons]="{selected: 'edit'}">
+
+    <nui-wizard-step-v2 i18n-label label="First step">
+        <span i18n>This is the first step of our wizard</span>.
+        <ng-template wizardStepFooter>
+            <ng-container *ngTemplateOutlet="wizardNext"></ng-container>
+        </ng-template>
+    </nui-wizard-step-v2>
+
+    <nui-wizard-step-v2 i18n-label label="Second step" [icon]="'email'">
```
Blakesalot, line 10: maybe `defaultIcon` would be better since it's only shown when the step isn't active or complete?

packages/bits/demo/src/components/demo/wizard-v2/wizard-with-custom-icons/wizard-with-custom-icons-example.component.html
```diff
@@ -0,0 +1,40 @@
...
+
+    <nui-wizard-step-v2 i18n-label label="Final step">
+        <span i18n>We finally reached our last step</span>
+        <ng-template wizardStepFooter>
+            <ng-container *ngTemplateOutlet="wizardPrevious"></ng-container>
+        </ng-template>
+    </nui-wizard-step-v2>
+
+    <ng-template #wizardPrevious>
+        <button type="button" nuiWizardPrevious nui-button icon="caret-left">
```
Blakesalot, line 26:
```suggestion
        <button type="button" nuiWizardPrevious nui-button icon="caret-left" i18n>
```

packages/bits/demo/src/components/demo/wizard-v2/wizard-with-custom-icons/wizard-with-custom-icons-example.component.html
```diff
@@ -0,0 +1,40 @@
...
+    </ng-template>
+
+    <ng-template #wizardNext>
+        <button nuiWizardNext
+                type="button"
+                nui-button
+                displayStyle="primary"
+                icon="caret-right"
+                iconRight="true"
+        >Next</button>
```
Blakesalot, line 38: Missing i18n
```suggestion
                i18n>
        Next</button>
```

packages/bits/demo/src/components/demo/wizard-v2/wizard-custom-layout/wizard-custom.component.less
Blakesalot: sounds good. yeah, a separate ticket works. do you think we should we remove these styles in the meantime?

dmytrofirsov: we'll see what we can do with this and all the rest in the next sprint, I really doubt we'll be able to complete it in the scope of the current one.

packages/bits/demo/src/components/demo/wizard-v2/wizard-with-custom-icons/wizard-with-custom-icons-example.component.html
dmytrofirsov: I think we'll take the next sprint to think the api through a bit better. This input, no matter how we call it, immediately makes me think of why are we allowing to change exactly this state out of 4 possible states. During our call we assumed this may be the only state consumers would want to change for the step, though, just like every assumption, this one feels a bit shaky, because our guess can be wrong, and then later we'll need to expand, or (the worst case) change the api which is not desirable.

dmytrofirsov:
Thank you @lukasholcik , @Blakesalot for your contributions and reviews. We'll take another iteration on this one in the next sprint and consider all of your comments left unanswered.

packages/bits/demo/src/components/demo/wizard-v2/wizard-with-custom-icons/wizard-with-custom-icons-example.component.html
Blakesalot: I like your thinking on this. ❤️

Blakesalot: Another thing to think about is whether we really need to have this little piece that's customizable if it's already possible to customize the entire wizard (see custom layout example). As you mentioned in the refinements meeting we don't really have any data on who will want to customize what. As a resultl, maybe we could even postpone this feature until we know what consumers will want to do?

packages/bits/src/constants/wizard.constans.ts
```diff
@@ -0,0 +1,11 @@
+import {InjectionToken} from "@angular/core";
+import {IWizardIcons} from "../lib/wizard-v2/wizard.directive";
+
+export const WIZARD_ICONS_PRESET_TOKEN = new InjectionToken<Partial<IWizardIcons>>("WIZARD_ICONS_PRESET");
+
+export const WIZARD_ICONS_PRESET: IWizardIcons = {
+    initial: "step",
+    visited: "step-complete",
+    selected: "step-active",
```
Blakesalot, line 9: Maybe we could use the same vernacular as the icon names?

```suggestion
    complete: "step-complete",
    active: "step-active",
```

packages/bits/src/constants/wizard.constans.ts
```diff
@@ -0,0 +1,11 @@
+import {InjectionToken} from "@angular/core";
+import {IWizardIcons} from "../lib/wizard-v2/wizard.directive";
+
+export const WIZARD_ICONS_PRESET_TOKEN = new InjectionToken<Partial<IWizardIcons>>("WIZARD_ICONS_PRESET");
+
+export const WIZARD_ICONS_PRESET: IWizardIcons = {
```
Blakesalot, line 6: Maybe this?

```suggestion
export const WIZARD_ICON_BY_STEP_STATE_PRESET: IWizardIcons = {
```

packages/bits/src/constants/wizard.constans.ts
```diff
@@ -0,0 +1,11 @@
+import {InjectionToken} from "@angular/core";
+import {IWizardIcons} from "../lib/wizard-v2/wizard.directive";
+
+export const WIZARD_ICONS_PRESET_TOKEN = new InjectionToken<Partial<IWizardIcons>>("WIZARD_ICONS_PRESET");
```
Blakesalot, line 4: Maybe this?

```suggestion
export const WIZARD_ICON_BY_STEP_STATE_PRESET_TOKEN = new InjectionToken<Partial<IWizardIcons>>("WIZARD_ICONS_PRESET");
```

packages/bits/src/lib/wizard-v2/wizard-horizontal/wizard-horizontal.component.html
```diff
@@ -13,16 +13,19 @@
...
             [optional]="step.optional"
             [errorMessage]="step.errorMessage"
-            [disableRipple]="disableRipple">
+            [disableRipple]="disableRipple"
+            [stepIcon]="step.icon"
+            [customIcons]="customIcons">
         </wizard-step-header>
-        <div *ngIf="!isLast" class="nui-wizard-horizontal-line"></div>
-    </ng-container>
+        <div *ngIf="!isLast" class="nui-wizard-horizontal-line" [ngClass]="{'nui-wizard-horizontal-line-completed': step.completed}"></div>
```
Blakesalot, line 27: I think a proper BEM class name would be something like the following. In addition to the underscores indicating the "element" part of BEM, I've added a name change suggestion. "connector" since it makes it more obvious when looking at the code what it's for, and removed the "completed" since the implementation detail for when to show it doesn't need to be embedded in the name.

```suggestion
        <div *ngIf="!isLast" class="nui-wizard-horizontal-line" [ngClass]="{'nui-wizard-horizontal__step-connector-line': step.completed}"></div>
```

packages/bits/src/lib/wizard-v2/wizard-horizontal/wizard-horizontal.component.ts
```diff
@@ -13,7 +13,7 @@ import { WizardDirective } from "../wizard.directive";
     host: {
         "class": "nui-wizard-horizontal-layout",
         "[class.nui-wizard-label-position-end]": "labelPosition == 'end'",
-        "[class.nui-wizard-label-position-bottom]": "labelPosition == 'bottom'",
+        "[class.nui-wizard-label-position-top]": "labelPosition == 'top'",
```
Blakesalot, line 16: For BEM, these should really be: `nui-wizard--step-label-position-end` and `nui-wizard--step-label-position-top`.

But, I don't understand why we're applying a class to the entire wizard host just for the step labels.

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.html
```diff
@@ -1,19 +1,8 @@
...
-            </span>
-        </ng-container>
-    </div>
-</div>
-<div class="nui-step-label"
-     [class.nui-step-label-active]="active"
-     [class.nui-step-label-selected]="selected"
-     [class.nui-step-label-error]="state == 'error'">
+<div class="nui-wizard-step-label"
+     [class.nui-wizard-step-header__label--active]="completed"
```
Blakesalot, line 3: Shouldn't this be "active"?
```suggestion
     [class.nui-wizard-step-header__label--active]="active"
```

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.html
```diff
@@ -1,19 +1,8 @@
 <div class="nui-step-header-ripple"></div>
```
Blakesalot, line 1:
```suggestion
<div class="nui-wizard-step-header__ripple"></div>
```

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.html
```diff
@@ -1,19 +1,8 @@
...
-                <nui-icon *ngSwitchDefault icon="{{_getDefaultIconForState(state)}}"></nui-icon>
-            </span>
-        </ng-container>
-    </div>
-</div>
-<div class="nui-step-label"
-     [class.nui-step-label-active]="active"
-     [class.nui-step-label-selected]="selected"
-     [class.nui-step-label-error]="state == 'error'">
+<div class="nui-wizard-step-label"
```
Blakesalot, line 2:
```suggestion
<div class="nui-wizard-step-header__label"
```

packages/bits/src/lib/wizard-v2/wizard-horizontal/wizard-horizontal.component.ts
Blakesalot, line 16: `==` --> `===`

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.html
```diff
@@ -1,19 +1,8 @@
...
+     [class.nui-wizard-step-header__label--active]="completed"
+     [class.nui-wizard-step-header__label--selected]="selected"
+     [class.nui-wizard-step-header__label--error]="state == 'error'">
```
Blakesalot, line 5:
```suggestion
     [class.nui-wizard-step-header__label--error]="state === 'error'">
```

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.html
```diff
@@ -1,19 +1,8 @@
...
 
     <!-- If there is a label template, use it. -->
     <ng-container *ngIf="_templateLabel()" [ngTemplateOutlet]="_templateLabel()!.template">
```
Blakesalot, line 8: I'm not sure what the `!` means in `"_templateLabel()!.template"`. Is this a syntax I'm not familiar with?

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.html
```diff
@@ -25,3 +14,14 @@
     <div class="nui-step-optional" *ngIf="optional && state != 'error'">{{optionalLabel}}</div>
```
Blakesalot, line 14:
```suggestion
    <div class="nui-wizard-step-header__label--optional" *ngIf="optional && state != 'error'">{{optionalLabel}}</div>
```

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.html
```diff
@@ -25,3 +14,14 @@
...
     <div class="nui-step-sub-label-error" *ngIf="state == 'error'">{{errorMessage}}</div>
 </div>
+
+<div class="nui-step-icon-state-{{state}} nui-wizard-step-icon"
```
Blakesalot, line 18:
```suggestion
<div class="nui-wizard-step-header__icon--{{state}} nui-wizard-step-header__icon"
```

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.html
```diff
@@ -25,3 +14,14 @@
...
+     [class.nui-wizard-step-header__icon--selected]="selected">
```
Blakesalot, line 19: This looks like it could be duplicate functionality with what's happening on line 18 with the `state` interpolation. Or, maybe the `selected` state styling logic could be merged with the styling logic for the other states somehow?

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.html
```diff
@@ -25,3 +14,14 @@
...
+    <div class="nui-step-icon-content">
```
Blakesalot, line 20:
```suggestion
    <div class="nui-wizard-step-header__icon-content">
```

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.html
```diff
@@ -25,3 +14,14 @@
...
+        <span class="nui-icon-container">
```
Blakesalot, line 21:
```suggestion
        <span class="nui-wizard-step-header__icon-container">
```

Blakesalot, line 21: It seems like the container class is inside the content class instead of the other way around. I would expect it to be:

```
<div class="container"> 
    <span class="content"></span>
</div>
```

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.ts
```diff
@@ -1,30 +1,51 @@
...
+    SimpleChanges,
+    ViewEncapsulation,
 } from "@angular/core";
 
 import {WizardStepLabelDirective} from "../wizard-step-label.directive";
+import {IWizardIcons} from "../wizard.directive";
+import {WIZARD_ICONS_PRESET, WIZARD_ICONS_PRESET_TOKEN} from "../../../constants";
 
 @Component({
     selector: "wizard-step-header",
```
Blakesalot, line 22:
```suggestion
    selector: "nui-wizard-step-header",
```

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.less
```diff
@@ -10,14 +10,20 @@
...
     .setCssVariable(color, nui-color-text-default);
-}
+    width: @wizard-label-header-width;
+
+    &.nui-wizard-step-header--completed {
+        .setCssVariable(color, nui-color-text-link);
 
-.nui-wizard-step-header-completed {
-    .setCssVariable(color, nui-color-text-link);
+        .nui-wizard-horizontal-line {
```
Blakesalot, line 18: This class is used in the wizard-horizontal template, so I think it should be moved to the less file for that component.

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.ts
```diff
@@ -1,30 +1,51 @@
...
     styleUrls: ["wizard-step-header.component.less"],
     host: {
         "class": "nui-wizard-step-header mat-focus-indicator",
-        "[class.nui-wizard-step-header-selected]": "selected",
-        "[class.nui-wizard-step-header-active]": "active",
-        "[class.nui-wizard-step-header-optional]": "optional",
-        "[class.nui-wizard-step-header-completed]": "state == 'done'",
+        "[class.nui-wizard-step-header--selected]": "selected",
+        "[class.nui-wizard-step-header--optional]": "optional",
+        "[class.nui-wizard-step-header--completed]": "state == 'done'",
```
Blakesalot, line 29:
```suggestion
        "[class.nui-wizard-step-header--completed]": "state === 'done'",
```

packages/bits/src/lib/wizard-v2/wizard.directive.ts
```diff
@@ -21,6 +27,15 @@ import { distinctUntilChanged, startWith, takeUntil } from "rxjs/operators";
 import { WizardStepV2Component } from "./wizard-step/wizard-step.component";
 import { IWizardState } from "./types";
 
+export interface IWizardIcons {
```
Blakesalot, line 30: Maybe this?

```suggestion
export interface IWizardIconsByStepState {
```

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.ts
```diff
@@ -101,15 +134,32 @@ export class WizardStepHeaderComponent extends CdkStepHeader implements AfterVie
         return returnedState;
     }
 
+    private getIconsForStates(icons: IWizardIcons, customIcons: IWizardIcons): IWizardIcons {
+        const stepIcons = icons;
+        if (customIcons) {
+            Object.keys(customIcons).forEach(icon => {
```
Blakesalot, line 140:
```suggestion
            Object.keys(customIcons).forEach(state => {
```

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.ts
```diff
@@ -40,20 +61,33 @@ export class WizardStepHeaderComponent extends CdkStepHeader implements AfterVie
...
         super(_elementRef);
+        if (this.injectedIcons) {
+            this.icons = this.getIconsForStates(this.icons, this.injectedIcons);
+        }
+
+    }
+
+    ngOnChanges(changes: SimpleChanges): void {
+        if (changes.customIcons) {
+            this.icons = this.getIconsForStates(this.icons, changes.customIcons.currentValue);
```
Blakesalot, line 89: We don't really need to pass `this.icons` since we can reference `this.icons` in the method implementation itself.

Blakesalot, line 89: maybe this?

```suggestion
            this.icons = this.getIconsByState(this.icons, changes.customIcons.currentValue);
```

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.ts
```diff
@@ -101,15 +134,32 @@ export class WizardStepHeaderComponent extends CdkStepHeader implements AfterVie
...
+            this.iconColor = "gray";
+        } else if (state === "done") {
+            returnedState = this.icons.visited;
+            this.iconColor = "primary-blue";
+        } else if (state === "edit") {
+            returnedState = this.icons.selected;
+            this.iconColor = "black";
+        } else if (state === "error") {
+            returnedState = this.icons.error;
+            this.iconColor = "red";
```
Blakesalot, line 162: I don't see where iconColor is being used. Besides that, though, we should avoid setting it here because it's an unexpected side-effect of calling _getDefaultIconForState.

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.ts
```diff
@@ -101,15 +134,32 @@ export class WizardStepHeaderComponent extends CdkStepHeader implements AfterVie
...
         let returnedState: StepState = state;
 
-        switch (state) {
-            case "number":
-            case "done": returnedState = "step-complete"; break;
-            case "edit": returnedState = "edit"; break;
-            case "error": returnedState = "warning"; break;
-            default:
+        if (state === "number") {
+            returnedState = this.stepIcon || this.icons.initial;
```
Blakesalot, line 152: I think it would be better to incorporate the value of `this.stepIcon` directly into `this.icons` to have a single source of truth for the icons.

packages/bits/src/lib/wizard-v2/wizard.component.less
```diff
@@ -34,31 +34,42 @@
...
-    border-top-width: @wizard-line-width;
+    border-top-width: @nui-line-medium;
     border-top-style: solid;
     .setCssVariable(border-color, nui-color-line-default);
 
-    flex: auto;
-    height: 0;
-    margin: 0 (@wizard-line-gap - @wizard-side-gap);
-    min-width: (@wizard-line-gap + @wizard-side-gap);
+    &.nui-wizard-horizontal-line-completed{
```
Blakesalot, line 46: This class and associated styling should live in the less file for wizard-horizontal

packages/bits/src/lib/wizard-v2/wizard.component.less
```diff
@@ -34,31 +34,42 @@
     }
 }
 
+.nui-step-header-wrapper{
+    position: relative;
+}
+
 .nui-wizard-horizontal-line {
```
Blakesalot, line 41: All of the horizontal-specific styles should live in the wizard-horizontal less file, and all of the vertical-specific files should live in the wizard-vertical less file.

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.ts
```diff
@@ -101,15 +134,32 @@ export class WizardStepHeaderComponent extends CdkStepHeader implements AfterVie
...
+            Object.keys(customIcons).forEach(icon => {
+                stepIcons[icon] = customIcons[icon];
+            });
+        }
+
+        return stepIcons
+    }
+
     _getDefaultIconForState(state: StepState): string {
         let returnedState: StepState = state;
```
Blakesalot, line 149:
```suggestion
        let icon: string;
```

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.ts
```diff
@@ -101,15 +134,32 @@ export class WizardStepHeaderComponent extends CdkStepHeader implements AfterVie
...
+        if (customIcons) {
+            Object.keys(customIcons).forEach(icon => {
+                stepIcons[icon] = customIcons[icon];
+            });
+        }
+
+        return stepIcons
+    }
+
     _getDefaultIconForState(state: StepState): string {
```
Blakesalot, line 148:
```suggestion
    getIcon(state: StepState): string {
```

packages/bits/src/lib/wizard-v2/wizard.directive.ts
```diff
@@ -116,6 +135,22 @@ export class WizardDirective extends CdkStepper implements OnChanges, AfterConte
         super.ngOnDestroy();
     }
 
+    public getStepState(i: number, state: StepState = STEP_STATE.NUMBER): StepState {
```
Blakesalot, line 138:
```suggestion
    public getStepState(stepIndex: number, state: StepState = STEP_STATE.NUMBER): StepState {
```

Blakesalot, line 138: I'm not sure it makes sense to have a state argument here. Can we just return `STEP_STATE.NUMBER` regardless at the bottom of the method?

packages/bits/src/lib/wizard-v2/wizard.directive.ts
```diff
@@ -116,6 +135,22 @@ export class WizardDirective extends CdkStepper implements OnChanges, AfterConte
...
+
+        if (step._showError && step.hasError && !isSelected) {
+            return STEP_STATE.ERROR;
+        } else if (isSelected) {
+            return STEP_STATE.EDIT
+        } else if (steps[i].completed) {
+            return STEP_STATE.DONE
+        } else {
+            return state;
+        }
```
Blakesalot, line 151: Maybe we don't need the `else` here:

```suggestion
        }
        
        return STEP_STATE.NUMBER;
```

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.ts
```diff
@@ -65,54 +101,62 @@ export class WizardStepHeaderComponent extends CdkStepHeader implements AfterVie
...
+        }
+        
+        if (this.stepState === "edit") {
+            return {
+                icon: this.wizardConfig.stepStateIcons?.icons?.active,
+                color: this.wizardConfig.stepStateIcons?.colors?.active,
+            };
+        }
+        
+        if (this.stepState === "error") {
```
DmytroPutrin, line 141: nit: mb better use here STEP_STATE instead string?

DmytroPutrin (dismissed):

packages/bits/demo/src/components/demo/wizard-v2/wizard-docs/wizard-docs.component.html
```diff
@@ -62,9 +62,27 @@ <h3>Async Form Validation</h3>
 </nui-example-wrapper>
 
 
-<h3>Wizard with custom icons</h3>
+<h3>Wizard with custom step state icons</h3>
 <p>
-    You are able to add custom icons to wizard.
+    The wizard step can be in four different states: <code>initial</code>, <code>selected</code>, <code>completed</code>, and <code>error</code>.
+    For each of these steps there is a corresponding icon in the header for a certain step, which dynamically changes along with the step state change.
+    It is possible to override the default set of step state icons and their colors, and there are two way of doing that.
```
Blakesalot, line 69:
```suggestion
    It is possible to override the default set of step state icons and their colors, and there are two ways of doing that.
```

packages/bits/demo/src/components/demo/wizard-v2/wizard-docs/wizard-docs.component.html
```diff
@@ -62,9 +62,27 @@ <h3>Async Form Validation</h3>
...
+</p>
+
+<ol>
+    <li>Global</li>
+    <li>Step Specific</li>
+</ol>
+
+<p><u>Global</u></p>
+<p>
+    Nova provides you with the <code>WIZARD_CONFIG</code> injection token, which allows to override the default wizard settings in a global context. For instance,
```
Blakesalot, line 79:
```suggestion
    Nova provides you with the <code>WIZARD_CONFIG</code> injection token, which allows you to override the default wizard settings in a global context. For instance, 
```

packages/bits/demo/src/components/demo/wizard-v2/wizard-docs/wizard-docs.component.html
```diff
@@ -62,9 +62,27 @@ <h3>Async Form Validation</h3>
...
+
+<ol>
+    <li>Global</li>
+    <li>Step Specific</li>
+</ol>
+
+<p><u>Global</u></p>
+<p>
+    Nova provides you with the <code>WIZARD_CONFIG</code> injection token, which allows to override the default wizard settings in a global context. For instance, 
+    set the necessary icons settings with this token on a module level to make all wizards within the module to receive the configuration you've set.
```
Blakesalot, line 80:
```suggestion
    set the necessary icons settings with this token on a module level to allow all wizards within a module to receive the same icon configuration.
```

packages/bits/demo/src/components/demo/wizard-v2/wizard-dynamic/wizard-dynamic.example.component.ts
```diff
@@ -1,6 +1,6 @@
 import {AfterViewInit, Component, TemplateRef, ViewChild} from "@angular/core";
 
-export interface IWizardStepConfig {
+export interface IWizardStepV2Component {
```
Blakesalot, line 3: What's the reason for this rename? To me this name sounds like a nova library interface instead of a demo consumer-side interface.

packages/bits/demo/src/components/demo/wizard-v2/wizard-remove-step/wizard-remove-step-example.component.ts
```diff
@@ -5,14 +5,14 @@ import {
     ViewChild,
 } from "@angular/core";
 
-import {IWizardStepConfig } from "../wizard-dynamic/wizard-dynamic.example.component";
+import {IWizardStepV2Component } from "../wizard-dynamic/wizard-dynamic.example.component";
```
Blakesalot, line 8: I would be good to avoid using a mystery interface from another example. Can we duplicate the interface in this example, so that the code is self-contained?

packages/bits/demo/src/components/demo/wizard-v2/wizard-restore-state/wizard-restore-state.example.component.ts
```diff
@@ -4,7 +4,7 @@ import { DialogService, NuiDialogRef, WizardStepV2Component, IWizardState } from
 import isEqual from "lodash/isEqual";
 import { Subject } from "rxjs";
 import { takeUntil } from "rxjs/operators";
-import { IWizardStepConfig } from "../wizard-dynamic/wizard-dynamic.example.component";
+import { IWizardStepV2Component } from "../wizard-dynamic/wizard-dynamic.example.component";
```
Blakesalot, line 7: Same comment as above. Can we duplicate this interface in here to make the example code self-contained (instead of importing it from another example).

packages/bits/demo/src/components/demo/wizard-v2/wizard-dynamic/wizard-dynamic.example.component.ts
Blakesalot: Would it make sense to move it to the library? If so, I think `IWizardStepConfig` works as a name. If we keep it local to the examples, maybe just `IStepConfig` since the `Wizard` part is kind of redundant in this context.

packages/bits/demo/src/components/demo/wizard-v2/wizard-with-custom-icons/wizard-with-custom-icons-example.component.html
```diff
@@ -1,29 +1,39 @@
...
+        </div>
         <ng-template wizardStepFooter>
             <ng-container *ngTemplateOutlet="wizardNext"></ng-container>
         </ng-template>
     </nui-wizard-step-v2>
 
-    <nui-wizard-step-v2 i18n-label label="Second step" [icon]="'email'">
-        <span i18n>This is the second step of our wizard</span>.
+    <!--
+        See how we use the 'stepIconsConfig' input here to override the global icons configuration
```
Blakesalot, line 13: Helpful comment. Nice!

packages/bits/demo/src/components/demo/wizard-v2/wizard-with-custom-icons/wizard-with-custom-icons-example.component.html
```diff
@@ -1,29 +1,39 @@
...
+        for this particular wizard step
+    -->
+    <nui-wizard-step-v2 i18n-label label="Second step" [stepIconsConfig]="secondStepIconConfig">
```
Blakesalot, line 16: Maybe we could rename the input `iconsConfig` since it's already in a context where the `step` part is redundant?

packages/bits/demo/src/components/demo/wizard-v2/wizard-with-custom-icons/wizard-with-custom-icons-example.component.ts
```diff
@@ -1,6 +1,32 @@
 import { Component } from "@angular/core";
+import { IWizardConfig, IWizardStepStateIconConfig, WIZARD_CONFIG } from "@nova-ui/bits";
 @Component({
     selector: "nui-wizard-with-custom-icons-example",
     templateUrl: "./wizard-with-custom-icons-example.component.html",
+    providers: [
+        {
+            provide: WIZARD_CONFIG,
+            useValue: {
+                stepStateIcons: {
```
Blakesalot, line 10: Maybe `stepStateIconsConfig` for better clarity that this is a configuration object rather than a collection of icons?

packages/bits/demo/src/components/demo/wizard-v2/wizard-with-custom-icons/wizard-with-custom-icons-example.component.ts
```diff
@@ -1,6 +1,32 @@
 import { Component } from "@angular/core";
+import { IWizardConfig, IWizardStepStateIconConfig, WIZARD_CONFIG } from "@nova-ui/bits";
 @Component({
     selector: "nui-wizard-with-custom-icons-example",
     templateUrl: "./wizard-with-custom-icons-example.component.html",
+    providers: [
```
Blakesalot, line 6: Maybe a comment here (in addition to the one in the docs text) reminding consumers they can use this construction in an ngModule's `providers` property to propagate the wizard configuration to all wizards within a module.

packages/bits/src/lib/wizard-v2/types.ts
```diff
@@ -2,3 +2,21 @@
...
+    stepStateIcons: IWizardStepStateIconConfig;
+    [key: string]: any;
+}
+
+export interface IWizardStepStateIconConfig {
+    icons?: Partial<IWizardIconStates>;
+    colors?: Partial<IWizardIconStates>;
+}
+
+export interface IWizardIconStates {
```
Blakesalot, line 16: Maybe something more generic like `IWizardStateMap`

packages/bits/src/lib/wizard-v2/types.ts
```diff
@@ -2,3 +2,21 @@
...
+    active: string;
+    visited: string;
+    initial: string;
+    error: string;
+    [key: string]: string;
```
Blakesalot, line 21: maybe the type of the members could be genericized with `<T>` to allow a mapping of states to any type in the future?

packages/bits/src/lib/wizard-v2/wizard-horizontal/wizard-horizontal.component.html
```diff
@@ -13,17 +13,16 @@
             [attr.aria-label]="step.ariaLabel || null"
             [attr.aria-labelledby]="(!step.ariaLabel && step.ariaLabelledby) ? step.ariaLabelledby : null"
             [index]="i"
-            [state]="getStepState(i, step.state)"
+            [stepState]="getStepState(i, step.state)"
```
Blakesalot, line 16: What's the reason for the rename here? Maybe there's a reason that I'm not thinking of, but could we just keep calling it `state`?

packages/bits/src/lib/wizard-v2/wizard-horizontal/wizard-horizontal.component.html
```diff
@@ -13,17 +13,16 @@
...
+            [stepIconsConfig]="step.stepIconsConfig"
```
Blakesalot, line 17:
```suggestion
            [iconsConfig]="step.iconsConfig"
```

packages/bits/src/lib/wizard-v2/types.ts
```diff
@@ -2,3 +2,21 @@
...
+
+export interface IWizardConfig {
+    stepStateIcons: IWizardStepStateIconConfig;
+    [key: string]: any;
+}
+
+export interface IWizardStepStateIconConfig {
+    icons?: Partial<IWizardIconStates>;
+    colors?: Partial<IWizardIconStates>;
+}
```
Blakesalot, line 14: I feel like we should couple the icons with the colors so that each icon state is self-contained instead of having separated properties.

```
export interface IWizardStepStateConfig {
    iconsConfig?: Partial<IWizardStateMap<IIconConfig>>;
}

export interface IWizardStateMap<T> {
    active: T;
    visited: T;
    initial: T;
    error: T;
    [key: string]: T;
}

export interface IIconConfig {
    icon: string;
    color: string;
} 
```

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.ts
```diff
@@ -99,70 +101,62 @@ export class WizardStepHeaderComponent extends CdkStepHeader implements AfterVie
...
-        let returnedState: StepState = state;
-        switch (state) {
-            case "number":
-                returnedState = `${this.index + 1}`;
-                break;
-
-            case "edit":
-                returnedState = "create";
-                break;
+    public get currentStepStateIconConfig(): IWizardCurrentStepStatusIconConfig | undefined {
```
Blakesalot, line 118: It would be great to have a state to iconConfig map that gets updated whenever the config changes. That way the all these if checks could be replaced with something like this:

```
return this.stateIconConfigMap[this.state];
```

packages/bits/src/lib/wizard-v2/wizard.directive.ts
```diff
@@ -135,20 +124,25 @@ export class WizardDirective extends CdkStepper implements OnChanges, AfterConte
         super.ngOnDestroy();
     }
 
-    public getStepState(i: number, state: StepState = STEP_STATE.NUMBER): StepState {
+    public getStepState(i: number, stepState: StepState = STEP_STATE.NUMBER): StepState {
```
Blakesalot, line 127: I think I've asked this before, but I'm not sure why we pass in the stepState here.

packages/bits/demo/src/components/demo/wizard-v2/wizard-dynamic/wizard-dynamic.example.component.ts
dmytrofirsov: This is a consumer-defined interface, its only purpose is to make it easier to work with the code and make typescript happy. This particular interface name explicitly describes the configuration target. Consumers can certainly use a different name or decide not to use any interfaces at all. I don't think we should restrict them with any pre-defined interfaces here.

dmytrofirsov: I renamed the interface to **IWizardStepData**. Hope this sounds better now

packages/bits/demo/src/components/demo/wizard-v2/wizard-remove-step/wizard-remove-step-example.component.ts
dmytrofirsov: That's a good catch, maybe a leftover from previous commits. Fixed

packages/bits/demo/src/components/demo/wizard-v2/wizard-restore-state/wizard-restore-state.example.component.ts
dmytrofirsov: fixed

packages/bits/demo/src/components/demo/wizard-v2/wizard-with-custom-icons/wizard-with-custom-icons-example.component.html
dmytrofirsov: The `iconsConfig` does not fit because the step component does not have any icons. I see now that the `stepIconsConfig`, however, does not fit either because of the same reason. Maybe `stateIconsConfig` will be a better choice because icons are directly bound to the step state, rather than to the state component itself.

packages/bits/demo/src/components/demo/wizard-v2/wizard-with-custom-icons/wizard-with-custom-icons-example.component.ts
dmytrofirsov: It seems logical to me that what's inside the config is a config. I'm not sure we should explicitly add `config` to config keys if they're part of the... config :)

packages/bits/demo/src/components/demo/wizard-v2/wizard-with-custom-icons/wizard-with-custom-icons-example.component.ts
dmytrofirsov: I can certainly do that!

The question I'm asking myself here is should we really remind Angular developers how to use angular DI in our component examples docs twice? I'm guessing you were thinking of people copying and pasting our examples without reading the docs description? In this case, they'd have that comment with a reminder in the examples code. Is my understanding valid?

packages/bits/src/lib/wizard-v2/types.ts
dmytrofirsov: Sounds good

packages/bits/src/lib/wizard-v2/types.ts
dmytrofirsov: The angular cdk explicitly defines the type and enum for the step states as string values. I can't think of any other type anybody would want to act there as a state name/value.

image

We can always generalize it later when we ever have the demand for it, can't we?

packages/bits/src/lib/wizard-v2/wizard-horizontal/wizard-horizontal.component.html
dmytrofirsov: The reason of the rename is to avoid confusion since the step header component does not have its own state, it only stores the step's state.

packages/bits/src/lib/wizard-v2/types.ts
dmytrofirsov: It depends on how we look at what we want to configure there. What you suggest looks more like configuring the step state first and then specify what exactly to configure inside the step state configuration, while in my case we configure the icons config first and don't bother with the rest. I like to keep things separated, but I agree that my way is a bit more specific. On the other hand, with this approach, it is a bit easier to only set the icons or colors separately, and it seems to be a bit more obvious from the consumer's point of view: in case they're only interested in configuring state icons they just send the appropriate config.

Also, there are additional questions that arise when attempting to configure the step state in the way you suggest: do we configure it for the step, or for the header? Icons only make sense for the header, but their type depends on the step state information, so how should we set this config: for the step, or for the header? In case we choose neither of these and simply add it as a root config option then what's the difference from what's already done, but specifically for the icons? Moreover, the step state does not know anything about any icons, it is just a part of the step.

If we really want to think everything through and make the interface for the wizard config as expandable as possible, we should probably consider starting from this structure:

`
{
    .... <wizard global settings>
    step:
        header?: IWizardHeaderConfig;
        body?: IWizardBodyConfig;
        footer?: IWizardFooterConfig;
    .... <other configs>
}
`
And each of these interfaces will (or will not) contain other fields and their subsequent structures, which maybe will never be used by anyone ever.

That's the way of thinking I had. We have no idea what consumers will want to configure in the future, and it seemed logical to me to keep it simple and start from the minimum needed configuration structure for the wizard and later expand the interface by simply adding new optional configuration entries, which is never a breaking change. Later on, in case we have so many configuration options that they'll start to pile up, we'll want to segregate the interface, and we'll do that in a major version bump.

What's your opinion on this @lukasholcik, @Blakesalot ?

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.ts
dmytrofirsov: Do you mean something like this? This also works.
image

What is the benefit of setting up a separate map over conditionally using the existing wizard config?

packages/bits/src/lib/wizard-v2/wizard.directive.ts
dmytrofirsov: this is a leftover. thanks for capturing

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.ts
dmytrofirsov: For instance, if angular will ever change the name of the step status from `edit` to something else, this will bread (as well as the code snippet with conditionals though). The thing is I can utilize the STEP_STATE enum, which I actually forgot about. It is provided by angular and I can use it in my conditionals, while I don't think it is possible to make the enum values be the object keys. We could construct the map object from the enum though, but do you think this is worth the effort?

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.ts
dmytrofirsov: that's a good point

packages/bits/src/lib/wizard-v2/types.ts
Blakesalot: Well, we're using the state map to map a state to a configuration item in our current case. The configuration items we're currently dealing with happen to be strings, but they could just as well be numbers, booleans, or even objects.

packages/bits/demo/src/components/demo/wizard-v2/wizard-with-custom-icons/wizard-with-custom-icons-example.component.ts
Blakesalot: In Alpha's experience with user testing for nova charts, we observed people habitually copying/pasting code without looking at the text at all. So yeah, it's kind of a fail-safe to make sure they know the intended use case for the injection token. Maybe less experienced developers could benefit from it. Just a thought. Not a required change. :)

packages/bits/demo/src/components/demo/wizard-v2/wizard-with-custom-icons/wizard-with-custom-icons-example.component.ts
Blakesalot: okay

packages/bits/demo/src/components/demo/wizard-v2/wizard-dynamic/wizard-dynamic.example.component.ts
Blakesalot: that does sound better :)

packages/bits/src/lib/wizard-v2/wizard.directive.ts
```diff
@@ -116,6 +124,26 @@ export class WizardDirective extends CdkStepper implements OnChanges, AfterConte
         super.ngOnDestroy();
     }
 
+    public getStepState(i: number): StepState {
```
lukasholcik, line 127: This method is quite large and is called from the template in every wizard implementation. I'd consider reworking it in a way that returns static value instead of calculating it every time. (Including operations like `toArray` and linear search of `indexOf`)

packages/bits/src/lib/wizard-v2/wizard-vertical/wizard-vertical.component.html
```diff
@@ -1,5 +1,5 @@
 <div class="nui-step" *ngFor="let step of steps; let i = index; let isLast = last">
-    <wizard-step-header class="nui-wizard-vertical-header"
+    <nui-wizard-step-header class="nui-wizard-vertical-header"
```
lukasholcik, line 2: according to BEM, it should be `nui-wizard-vertical__header`

dmytrofirsov: It was decided to stop working on the vertical wizard since there is no use case and we have no specs for it. We'll be reworking this type of wizard once there is a demand and clear requirements. I removed this component from public apis to not confuse anyone.

DmytroPutrin (approved):

packages/bits/src/lib/wizard-v2/types.ts
```diff
@@ -1,22 +1,17 @@
 
+export type WizardIconStates = "active" | "visited" | "initial" | "error" | string;
```
Blakesalot, line 2: Can we call this just `WizardState`?

packages/bits/src/lib/wizard-v2/types.ts
```diff
@@ -1,22 +1,17 @@
...
+export type WizardStepStateConfig = Record<WizardIconStates, IWizardStepStateConfig>;
```
Blakesalot, line 3: Can we call it `WizardStateConfigMap` perhaps?

packages/bits/src/lib/wizard-v2/types.ts
```diff
@@ -1,22 +1,17 @@
...
-export interface IWizardIconStates {
-    active: string;
-    visited: string;
-    initial: string;
-    error: string;
-    [key: string]: string;
+export interface IWizardStepStateConfig {
+    icon?: string;
+    color?: string;
 }
```
Blakesalot: I was thinking it would look like this:

```
export interface IWizardStepStateConfig {
    icon: IIconConfig;
}

export interface IIconConfig{
    icon?: string;
    color?: string;
}
```

That way it's obvious what the color property if for. But, I would settle for this too.

```
export interface IWizardStepStateConfig {
    icon?: string;
    iconColor?: string;
}
```

packages/bits/src/lib/wizard-v2/types.ts
Blakesalot: An, never mind. I like your idea better on this one.

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.ts
```diff
@@ -115,48 +116,16 @@ export class WizardStepHeaderComponent extends CdkStepHeader implements AfterVie
...
-            colors: {
-                ...this.wizardConfig.stepStateIcons?.colors,
-                ...iconConfig?.colors,
-            },
+    private createStepStateConfigMap() {
+        this.stepStateConfigMap = {
+            [STEP_STATE.NUMBER]: this.wizardConfig.stepState?.initial,
+            [STEP_STATE.DONE]: this.wizardConfig.stepState?.visited,
+            [STEP_STATE.EDIT]: this.wizardConfig.stepState?.active,
+            [STEP_STATE.ERROR]: this.wizardConfig.stepState?.error,
```
Blakesalot, line 128: I like it. Nice job!

Vita-Karas (approved):

ivankozlovskyi (approved):

PR #314, opened 2021-05-31
menu keyboard nav

dmytrofirsov:


ivankozlovskyi (approved):

packages/bits/spec/components/menu/menu.atom.ts
```diff
@@ -13,6 +14,9 @@ import { MenuItemAtom } from "./menu-item.atom";
 
 export class MenuAtom extends Atom {
     public static CSS_CLASS = "nui-menu";
+
+    public static getAllMenuItems = () => element.all(by.css(".nui-menu-item:not(.nui-menu-item--header"));
```
Blakesalot, line 18: I'm wondering if it might be more user-friendly to return an array of MenuItemAtom's instead of an ElementFinderArray.

packages/bits/spec/components/menu/menu.atom.ts
```diff
@@ -36,6 +40,7 @@ export class MenuAtom extends Atom {
 
     public getMenuItemByIndex = (idx: number): MenuItemAtom => new MenuItemAtom(this.getItemElements().get(idx));
 
+    // DO NOT USE. The following method is broken and will be fixed in the scope of NUI-6105
```
Blakesalot, line 43: It may be a good idea to change this comment to a jsdoc-style comment so it will show up in an IDE's auto-complete tooltip if someone tries to use it.

packages/bits/src/lib/menu/menu-key-control.service.ts
```diff
@@ -91,27 +102,35 @@ export class MenuKeyControlService implements OnDestroy {
...
         }
-        if (event.keyCode === PAGE_UP) {
+        if (event.code === KEYBOARD_CODE.PAGE_UP || event.code === KEYBOARD_CODE.HOME  || (event.metaKey && event.code === KEYBOARD_CODE.ARROW_UP)) {
             event.preventDefault();
             this.keyboardEventsManager.onKeydown(event);
             this.keyboardEventsManager.setFirstItemActive();
+            this.announceCurrentItem();
         }
-        if (event.keyCode === PAGE_DOWN) {
+        if (event.code === KEYBOARD_CODE.PAGE_DOWN || event.code === KEYBOARD_CODE.END  || (event.metaKey && event.code === KEYBOARD_CODE.ARROW_DOWN)) {
```
Blakesalot, line 116: I wonder why PAGE_UP and PAGE_DOWN take the user all the way to the top or bottom instead of going page by page. I know this is pre-existing behavior, but maybe you know why?

dmytrofirsov: I think if it is here it was discussed with UX. I doubt there are any docs for that though. According to the w3c requirements, PAGE UP and PAGE DOWN keys are not required to participate in the navigation flow^ however, if it is here, it was discussed. I can also tell there was no such a requirement to navigate page by page for the component.

packages/bits/spec/components/menu/menu.atom.ts
dmytrofirsov: The broken method does exactly that and fails to deliver. So, until we know the root cause of the problem, I played safe and worked with ElementFinderArray until we fix the broken method which delivers menu item atoms.

Blakesalot: okay

packages/bits/src/lib/menu/menu-key-control.service.ts
Blakesalot: okay

ivankozlovskyi (approved):

dmytrognatyshyn (approved):

PR #315, opened 2021-05-31
Bugfix/NUI-4906 Tests review and improvements for expander component

elenates:
- Deleted a few e2e tests which were duplicated by visual tests.
- Added a visual test instead of e2e test to reduce time of test duration.
- Removed browser refresh before every test.
- Added return to initial state for tests which change the state (for example, close the expander after it was expanded in a test).
- Created a separated testing page for the test as e2e test were done on docs page.

### DIFF
--- a/packages/bits/demo/src/components/demo/expander/expander-test/expander-test.component.html
+++ b/packages/bits/demo/src/components/demo/expander/expander-test/expander-test.component.html
@@ -13,14 +13,10 @@
             <hr />
         </div>
         <div class="col">
-            <nui-expander-initially-expanded-example>
-                <nui-expander id="nui-demo-expander-initially-expanded" header="Advanced Options"></nui-expander>
-            </nui-expander-initially-expanded-example>
+            <nui-expander-initially-expanded-example></nui-expander-initially-expanded-example>
         </div>
         <div class="col">
-            <nui-expander-open-change-example>
-                <nui-expander id="nui-demo-expander-open-change" header="Advanced options"></nui-expander>
-            </nui-expander-open-change-example>
+            <nui-expander-open-change-example></nui-expander-open-change-example>
         </div>
     </div>
 </div>
--- a/packages/bits/demo/src/components/demo/expander/expander-test/expander-test.component.html
+++ b/packages/bits/demo/src/components/demo/expander/expander-test/expander-test.component.html
@@ -1,0 +1,37 @@
+<div class="container">
+    <div class="row">
+        <div class="col">
+            <nui-expander id="nui-demo-expander-basic">
+                <p>Content is here</p>
+            </nui-expander>
+            <hr />
+        </div>
+        <div class="col">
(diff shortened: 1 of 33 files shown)


ivankozlovskyi (approved):

packages/bits/demo/src/components/demo/expander/expander-test/expander-test.component.html
```diff
@@ -0,0 +1,37 @@
+<div class="container">
+    <div class="row">
+        <div class="col">
+            <nui-expander id="nui-demo-expander-basic">
```
Blakesalot, line 4: I think the same id's in the original docs examples that these were pulled from can safely be removed since they were only there for testing purposes. Removing the unneeded id's would help clean up the examples a little bit. This one for example: 

https://github.com/solarwinds/nova/blob/f85ab1e22d20ac96cf786ec8be67ed159fffa6df/packages/bits/demo/src/components/demo/expander/expander-basic/expander-basic.example.component.html#L1

Blakesalot: It might be good to do a quick search in the code base for each of them to make sure I'm right.

packages/bits/demo/src/components/demo/expander/expander-test/expander-test.component.html
```diff
@@ -0,0 +1,26 @@
...
+            <nui-expander-text-and-icon-example></nui-expander-text-and-icon-example>
+            <hr />
+        </div>
+        <div class="col">
+            <nui-expander-custom-header-example></nui-expander-custom-header-example>
+            <hr />
+        </div>
+        <div class="col">
+            <nui-expander-initially-expanded-example>
+                <nui-expander id="nui-demo-expander-initially-expanded" header="Advanced Options"></nui-expander>
```
Blakesalot, line 17: It seems like this line can be removed

packages/bits/demo/src/components/demo/expander/expander-test/expander-test.component.html
```diff
@@ -0,0 +1,26 @@
...
+            </nui-expander-initially-expanded-example>
+        </div>
+        <div class="col">
+            <nui-expander-open-change-example>
+                <nui-expander id="nui-demo-expander-open-change" header="Advanced options"></nui-expander>
```
Blakesalot, line 22: It seems like this line can be removed

packages/bits/demo/src/components/demo/expander/expander-test/expander-test.component.html
elenates: Thanks, removed.

packages/bits/demo/src/components/demo/expander/expander-test/expander-test.component.html
elenates: Thanks, removed.

Blakesalot (approved):

ivankozlovskyi (approved):

PR #316, opened 2021-06-01
combobox-v2 - active item now remains visible all the time during the…

dmytrofirsov:
… options filtering

### DIFF
--- a/packages/bits/src/lib/select-v2/combobox-v2/combobox-v2.component.ts
+++ b/packages/bits/src/lib/select-v2/combobox-v2/combobox-v2.component.ts
@@ -122,6 +122,8 @@ export class ComboboxV2Component extends BaseSelectV2 implements AfterContentIni
         // we check "selectedOptions" to be set per "value" again in "handleValueChange"
         this.optionsChanged().subscribe(() => {
             this.filterItems(this.inputValue.toString());
+            // We need the active option to always stay visible during the options filtering.
+            this.optionKeyControlService.scrollToOption({block: "start", behavior: "smooth"});
             this.cdRef.markForCheck();
         });
     }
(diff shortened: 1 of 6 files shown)


packages/bits/src/lib/select-v2/option-key-control.service.ts
```diff
@@ -45,6 +45,15 @@ export class OptionKeyControlService<T extends IOption> {
         this.keyboardEventsManager.skipPredicate(predicate);
     }
 
+    public scrollToOption(options: ScrollIntoViewOptions): void {
```
Blakesalot, line 48: It's just a suggestion, but I think `scrollToActiveItem` might be a better name for this method, and since it's newly public, it wouldn't be a breaking change to rename it.

dmytrofirsov: done

Vita-Karas (dismissed):

ivankozlovskyi (approved):

packages/bits/src/lib/select-v2/combobox-v2/combobox-v2.component.ts
```diff
@@ -122,6 +122,8 @@ export class ComboboxV2Component extends BaseSelectV2 implements AfterContentIni
         // we check "selectedOptions" to be set per "value" again in "handleValueChange"
         this.optionsChanged().subscribe(() => {
             this.filterItems(this.inputValue.toString());
+            // We need the active option to always stay visible during the options filtering.
+            this.optionKeyControlService.scrollToOption({block: "start", behavior: "smooth"});
```
Blakesalot: I know y'all are pressed for time, but do you have time to write a unit test that would be red before this fix and green after the fix?

dmytrofirsov: added e2e test to track the regression

ivankozlovskyi (approved):

dmytrognatyshyn (approved):

PR #317, opened 2021-06-02
bugfix/NUI-6064-Combobox-V2-preserve-selected-value-when-user-remove-value

Vita-Karas:


ivankozlovskyi (approved):

packages/bits/src/lib/select-v2/combobox-v2/combobox-v2.component.ts
```diff
@@ -126,26 +126,26 @@ export class ComboboxV2Component extends BaseSelectV2 implements AfterContentIni
...
 
-    public ngAfterViewInit() {
+    public ngAfterViewInit(): void {
         super.ngAfterViewInit();
         if (!this.multiselect) {
             this.dropdown.hide$.subscribe(() => {
                 const lastSelectedOption = this.getLastSelectedOption();
                 if (lastSelectedOption) {
-                    this.setInputValue(lastSelectedOption.viewValue);
+                    this.setInputValue(lastSelectedOption.value);
```
Blakesalot, line 135: It would be good to add a unit test that's red before this fix and green after the fix.

Vita-Karas: Added the e2e test

ivankozlovskyi (approved):

packages/bits/spec/components/combobox-v2/combobox-v2.e2e.ts
```diff
@@ -117,6 +117,21 @@ describe("USERCONTROL Combobox >", () => {
...
+            
+            it("should focus on the first item in dropdown, when removing item on backspace", async () => {
+                await comboboxError.click();
+                const option = await comboboxError.getOption(5);
+                await option.click();
+                
+                await expect(await comboboxError.getInputValue()).toEqual("Item 5");
+                await expect(await (await comboboxError.getOption(5)).isActive()).toBe(true);
+
+                browser.actions().doubleClick(comboboxError.input);
```
dmytrofirsov, line 129: might need to resolve the returned promise explicitly here with await

dmytrofirsov: surprisingly, doubleclick does not return promise for some reason :) ignore the comment above

dmytrofirsov (approved):

PR #321, opened 2021-06-07
Feature/NUI-4916 Tests review and improvements for progress component

elenates:
- Deleted e2e test duplicated in visual tests.
- Added one test for compact progress.
- Removed browser refresh before every test.
- Created a separated testing page for the test as e2e test were done on docs page.

### DIFF
--- a/packages/bits/spec/components/menu/menu.e2e.ts
+++ b/packages/bits/spec/components/menu/menu.e2e.ts
@@ -1,15 +1,8 @@
-import {
-    browser,
-    by,
-    ElementArrayFinder,
-    protractor,
-} from "protractor";
+import { browser, by, protractor } from "protractor";
 
 import { Atom } from "../../atom";
 import { CheckboxAtom } from "../../components/checkbox/checkbox.atom";
-import { SwitchAtom } from "../../components/switch/switch.atom";
 import { Helpers } from "../../helpers";
-
 import { MenuAtom } from "./menu.atom";
 
 describe("USERCONTROL Menu", () => {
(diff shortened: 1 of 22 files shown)


packages/bits/spec/components/menu/menu.e2e.ts
```diff
@@ -65,16 +68,13 @@ describe("USERCONTROL Menu", () => {
             });
 
             it("should NOT open and close menu by ENTER key if focused on toggle", async () => {
-                await menu.toggleMenu();
-                await menu.toggleMenu();
```
Blakesalot, line 69: It seems like these might be needed to ensure that the correct menu has focus before hitting the enter key.

packages/bits/spec/components/progress/progress.e2e.ts
```diff
@@ -35,6 +37,22 @@ describe("USERCONTROL progress", () => {
...
+    });
+
+    describe("compact progress", () => {
+        beforeEach(async () => {
+            await compactProgressBtn.click();
+        });
+
+        it("should show and hide progress bar when click button in compact progress", async () => {
+            expect(await compactProgress.isProgressBarDisplayed()).toBe(true);
+            expect(await compactProgress.hasClass("nui-progress--shown"));
```
Blakesalot, line 53: I'm wondering if these two queries are redundant? Do we need them both?

elenates: You are right, Blake, I will delete one of them, that is the same.

Blakesalot (approved):

madelyneh (approved):

ivankozlovskyi (approved):

PR #326, opened 2021-06-09
Feature/n UI 5923 add arrow nav for toolbar

DmytroPutrin:


ivankozlovskyi (dismissed):

Vita-Karas (dismissed):

packages/bits/src/lib/toolbar/toolbar-keyboard.service.ts
```diff
@@ -0,0 +1,100 @@
...
+    public onKeyDown(event: KeyboardEvent): void {
+        const { code } = event;
+
+        if (code === KEYBOARD_CODE.ARROW_LEFT || code === KEYBOARD_CODE.ARROW_RIGHT) {
+            event.preventDefault();
+            this.navigateByArrow(code);
+        }
+    }
+
+    private navigateByArrow(code: string): void {
```
Blakesalot, line 40: Can we used the KEYBOARD_CODE as the type instead of string?

Blakesalot, line 40: There's quite a bit of branching logic in this method, and it doesn't look like the unit tests we have cover most of the possibilities in here. It would be good to add tests to cover ~80% of the various conditions.

packages/bits/src/lib/toolbar/toolbar-keyboard.service.ts
```diff
@@ -0,0 +1,100 @@
...
+        if (nextIndex >= length && !this.moreBtn) {
+            nextIndex = 0;
+        }
+
+        const button: HTMLButtonElement = buttons[nextIndex];
+
+        button.focus();
+    }
+
+    private navigateFormMoreBtn(direction: number, buttons: HTMLButtonElement[]): void {
```
Blakesalot, line 76: Maybe we could capture the direction in an enum and use that instead of raw numbers?

```
export enum NavigationDirection {
    Left: 0,
    Right: 1,
}
```

Blakesalot, line 76: I'm not sure, but this may be a typo:

```suggestion
    private navigateFromMoreBtn(direction: number, buttons: HTMLButtonElement[]): void {
```

packages/bits/src/constants/keycode.constants.ts
```diff
@@ -23,6 +23,8 @@ export enum KEY_CODE {
 
 export enum KEYBOARD_CODE {
```
lukasholcik, line 24: There is a collection of key constants in the cdk library here: https://github.com/angular/components/blob/master/src/cdk/keycodes/keycodes.ts
I'd prefer to use that instead of a custom enum.

packages/bits/src/lib/toolbar/toolbar-keyboard.service.ts
```diff
@@ -0,0 +1,100 @@
...
+        if (this.dynamicContainer) {
+            this.moreBtn = this.dynamicContainer.querySelector(".nui-button.btn.menu-button");
+        }
+
+        if (menu) {
+            this.menuComponent = menu;
+        }
+    }
+
+    public disableFocusForMoreBtn(): void {
```
Blakesalot, line 25: This name would read better to me. Just a nit:

```suggestion
    public disableMoreBtnFocus(): void {
```

packages/bits/src/lib/toolbar/toolbar-keyboard.service.ts
lukasholcik: When using the CDK constants mentioned above, it would be just numbers.

packages/bits/src/lib/toolbar/toolbar.component.spec.ts
```diff
@@ -155,5 +158,36 @@ describe("components >", () => {
...
+            it("should call keyboardService 'disableFocusForMoreBtn' and 'initService' on ngAfterViewInit hook", () => {
+                const disableFocusForMoreBtnSpy = spyOn(component["keyboardService"], "disableFocusForMoreBtn");
+                const initServiceSpy = spyOn(component["keyboardService"], "initService");
+
+                component.ngAfterViewInit();
+                expect(disableFocusForMoreBtnSpy).toHaveBeenCalled();
+                expect(initServiceSpy).toHaveBeenCalled();
+            });
+
+            it("should call preventDefault", () => {
```
Blakesalot, line 185: It seems like this test might belong in the spec file for the service since that's where preventDefault is actually getting called from.

packages/bits/src/lib/toolbar/toolbar.component.html
```diff
@@ -3,10 +3,11 @@
      (containerResize)="moveToolbarItems()"
      #toolbarContainer>
     <div class="nui-toolbar-content__dynamic d-flex flex-row flex-shrink-0">
-        <ng-template ngFor let-commandGroup let-last="last" [ngForOf]="commandGroups">
-            <ng-template ngFor let-commandItem [ngForOf]="commandGroup.items">
+        <ng-template ngFor let-i="index" let-commandGroup let-last="last" [ngForOf]="commandGroups">
+            <ng-template ngFor let-commandItem let-j="index" [ngForOf]="commandGroup.items">
                 <button nui-button type="button"
-                        (click)="commandItem.actionDone.emit()"
+                        [tabindex]="(!i && !j) ? 0 : -1"
```
Blakesalot, line 9: Just a quick question. I'm not exactly sure what this is doing. Do you mind explaining?

packages/bits/src/constants/keycode.constants.ts
DmytroPutrin: keyCode was deprecated few years ago. That why I prefer to avoid using it.
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

packages/bits/src/lib/toolbar/toolbar.component.html
DmytroPutrin: as we decided don't put focus on all toolbar container, I just use firs element as a trap to catch focus. Then user could navigate by arrows. For rest of buttons I set tabindex to -1 to allow move from toolbar by pressing Tab. (!i && 1J) is not obvious and I could add comment for this if you wish. i - point to first group inside toolbar, j - to firs item inside this group.

packages/bits/src/lib/toolbar/toolbar.component.spec.ts
DmytroPutrin: agree, but here unfortunately, a bunch of imperative works with dom and I firstly decided to test service methods relatively via component. Because inside it I we have access to templete.

packages/bits/src/lib/toolbar/toolbar-keyboard.service.ts
DmytroPutrin: thanks

packages/bits/src/lib/toolbar/toolbar.component.html
Blakesalot: Thanks, Dima. I think an inline comment about that would indeed be helpful.

packages/bits/src/lib/toolbar/toolbar.component.spec.ts
Blakesalot: ok

packages/bits/src/lib/toolbar/toolbar-keyboard.service.ts
DmytroPutrin: added enum

packages/bits/src/lib/toolbar/toolbar-keyboard.service.ts
DmytroPutrin: the problem is that activeElement only readable property and I couldn't mock it. For testing this branching I need template, unfortunately, even toolbar template doesn't feet enough - debugElement doesn't have , moreButton, for instance. Honestly, I didn't find way how to test this lines of code correctly. By splitting test on two files  toolbar-keyboard.service.ts has approximately 80% line coverage. But it isn't good quality tests.

packages/bits/src/lib/toolbar/toolbar-keyboard.service.ts
```diff
@@ -2,6 +2,11 @@ import { Injectable } from "@angular/core";
 import { KEYBOARD_CODE } from "../../constants";
 import { MenuComponent } from "../menu";
 
+enum NavDirection {
```
Blakesalot, line 5: Thanks for adding the enum. Do you mind exporting it from [this](https://github.com/solarwinds/nova/blob/f1dbe170b14cc999ce5233b19844a093d43fc66e/packages/bits/src/lib/toolbar/public-api.ts) file and using the enum instead of a number in all the places where it can be used? For example in the service spec:

```
service["navigateFromMoreBtn"](NavDirection.Left, buttonsMock)
```

packages/bits/src/lib/toolbar/toolbar-keyboard.service.ts
```diff
@@ -2,6 +2,11 @@ import { Injectable } from "@angular/core";
...
+    LEFT = -1,
```
Blakesalot, line 6: Should this be `0`?

packages/bits/src/lib/toolbar/toolbar.component.html
```diff
@@ -5,6 +5,7 @@
     <div class="nui-toolbar-content__dynamic d-flex flex-row flex-shrink-0">
         <ng-template ngFor let-i="index" let-commandGroup let-last="last" [ngForOf]="commandGroups">
             <ng-template ngFor let-commandItem let-j="index" [ngForOf]="commandGroup.items">
+                <!-- (!i && !j) is used for set focusable state to firs element firs group -->
```
Blakesalot, line 8:
```suggestion
                <!-- (!i && !j) is used for set focusable state to first element in first group -->
```

packages/bits/src/lib/toolbar/toolbar-keyboard.service.ts
Blakesalot: And, one last thing, do you mind making the entries of the enum PascalCase? We've been trying to go with that for a convention. I'll see if it's documented somewhere. If not, I'll add it to the style guide. Thanks in advance.

e.g. `Left` instead of `LEFT`

packages/bits/src/lib/toolbar/toolbar-keyboard.service.ts
DmytroPutrin: I use it to get index for next element inside buttons array. Left - step back(-1) , current = 0, right - step forward(1);

packages/bits/src/lib/toolbar/toolbar-keyboard.service.ts
DmytroPutrin: Dima Firsov offer other solution, I applied it and ut.

packages/bits/src/lib/toolbar/toolbar-keyboard.service.ts
DmytroPutrin: remove it at all.

packages/bits/src/lib/toolbar/toolbar.component.html
DmytroPutrin: thanks

packages/bits/spec/components/toolbar/toolbar.e2e.ts
```diff
@@ -103,10 +103,10 @@ describe("USERCONTROL toolbar: ", () => {
 
         it("should toolbar resized and hide all items in menu", async () => {
             const width = "100";
-            await toolbar.popup.open();
             const toolbarItems = await toolbar.getAllVisibleItemsCount();
             await resizableToolbarTextboxAtom.acceptText(width);
             await browser.wait( async () => toolbarItems !== 0, 500);
+            await toolbar.popup.open();
```
Blakesalot, line 109: what was the reason this needed to be moved?

packages/bits/src/lib/toolbar/toolbar-keyboard.service.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+        btn.classList.add("testButton");
+        dynamicContainer.appendChild(btn);
+    }
+
+    container.appendChild(dynamicContainer);
+    fragment.appendChild(container);
+    document.body.appendChild(fragment);
+}
+
+createTestDomElements();
```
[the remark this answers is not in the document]
Blakesalot: proviso: I could be wrong about each suite using the same document instance, but cleanup in an afterAll is still a good practice for things like this.

packages/bits/src/lib/toolbar/toolbar-keyboard.service.ts
```diff
@@ -0,0 +1,47 @@
...
+    public onKeyDown(event: KeyboardEvent): void {
+        const { code } = event;
+
+        if (code === KEYBOARD_CODE.ARROW_LEFT || code === KEYBOARD_CODE.ARROW_RIGHT) {
+            event.preventDefault();
+            this.navigateByArrow(code);
+        }
+    }
+
+    private navigateByArrow(code: KEYBOARD_CODE): void {
```
Blakesalot, line 21: nice refactor!

packages/bits/spec/components/toolbar/toolbar.e2e.ts
DmytroPutrin: on resize trigger menu now lose focus. On this event we close menu and it isn't possible to reach menu items in visual test

packages/bits/src/lib/toolbar/toolbar.component.html
```diff
@@ -15,7 +18,11 @@
             </ng-template>
             <nui-divider *ngIf="!last || showMenu" class="flex-shrink-0" isVertical="true" size="extra-small"></nui-divider>
         </ng-template>
-        <nui-menu displayStyle="action" [title]="menuTitle" *ngIf="showMenu">
+        <nui-menu displayStyle="action"
+                  #menuComponent
+                  [title]="menuTitle"
+                  *ngIf="showMenu"
+                  (focusout)="menuComponent.popup.isOpen = false">
```
Blakesalot, line 25: I wonder if this behavior should be built into the menu instead of having to tell the menu "hey, your popup should be closed since you're no longer focused". @dmytrofirsov, @DmytroPutrin what do y'all think?

packages/bits/src/lib/toolbar/toolbar-keyboard.service.ts
```diff
@@ -0,0 +1,47 @@
+import { Injectable } from "@angular/core";
+import { KEYBOARD_CODE } from "../../constants";
+
+@Injectable()
```
Blakesalot, line 4: Does this need a `{ providedIn: "root" }` so that consumers don't need to mention it in their module providers arrays?

packages/bits/src/lib/toolbar/toolbar.component.html
DmytroPutrin: yes, that is hack. Menu has that functional, but not in all cases focusout fires in menu component. Why and exactly when I couldn't answer right now. But I will take a look in the end of the week.

packages/bits/spec/components/toolbar/toolbar.e2e.ts
Blakesalot: okay, thanks for the explanation

packages/bits/src/lib/toolbar/toolbar-keyboard.service.spec.ts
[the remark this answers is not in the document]
DmytroPutrin: got it, fixed.

packages/bits/src/lib/toolbar/toolbar-keyboard.service.ts
Blakesalot: nevermind, I just realized that each toolbar component has its very own instance of the service

DmytroPutrin: we inject this service in component scope, not module.

packages/bits/src/lib/toolbar/toolbar.component.html
Blakesalot: thanks 👍

packages/bits/src/lib/toolbar/toolbar.component.ts
```diff
@@ -54,6 +60,8 @@ export class ToolbarComponent implements AfterViewInit, OnDestroy {
     @ContentChildren(ToolbarItemComponent, {descendants: true})
     public items: QueryList<ToolbarItemComponent>;
 
+    @ViewChild("menuComponent") menuComponent: MenuComponent;
```
Blakesalot, line 63: This seems to be a duplicate member

packages/bits/src/lib/toolbar/toolbar.component.ts
```diff
@@ -183,4 +208,27 @@ export class ToolbarComponent implements AfterViewInit, OnDestroy {
     public get destructiveIsLastItem() {
         return this.groups.last.items.last.displayStyle !== ToolbarItemDisplayStyle.destructive;
     }
+
+    private subscribeToToolbarStepsChanges() {
+        this.toolbarButtons.changes
+            .pipe(takeUntil(this.destroy$))
```
Blakesalot, line 214: This seems to be missing from ngOnDestroy:

```
this.destroy$.next();
this.destroy$.complete();
```

packages/bits/src/lib/toolbar/toolbar.component.ts
```diff
@@ -183,4 +208,27 @@ export class ToolbarComponent implements AfterViewInit, OnDestroy {
...
+            .subscribe((buttons: QueryList<ButtonComponent>) => {
+                this.toolbarItems = buttons.toArray().slice().map(b => b["el"].nativeElement as HTMLElement);
+
+                if (this.menu && this.menu.menuToggle) {
+
+                    if (!buttons.length) {
+                        // In case all buttons are hidden within the Commands menu we want this menu to receive the focus
+                        this.menu.menuToggle.nativeElement.setAttribute("tabindex", "0");
+                    } else {
+                        // If at least one button is visible in the toolbar it should receive the focus fist upon navigating onto the toolbar
```
Blakesalot, line 224:
```suggestion
                        // If at least one button is visible in the toolbar it should receive the focus first upon navigating onto the toolbar
```

packages/bits/src/lib/toolbar/toolbar.component.ts
DmytroPutrin: thanks for notice

packages/bits/src/lib/toolbar/toolbar.component.html
Blakesalot: If you create a ticket, do you mind adding a note to it that this hack should be removed once it's fixed?

packages/bits/src/lib/toolbar/toolbar.component.ts
```diff
@@ -183,4 +208,27 @@ export class ToolbarComponent implements AfterViewInit, OnDestroy {
...
         return this.groups.last.items.last.displayStyle !== ToolbarItemDisplayStyle.destructive;
     }
+
+    private subscribeToToolbarStepsChanges(): void {
+        this.toolbarButtons.changes
+            .pipe(takeUntil(this.destroy$))
+            .subscribe((buttons: QueryList<ButtonComponent>) => {
+                this.toolbarItems = buttons.toArray().slice().map(b => b["el"].nativeElement as HTMLElement);
+
+                if (this.menu && this.menu.menuToggle) {
```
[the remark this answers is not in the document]
DmytroPutrin: your are right, fixed

packages/bits/src/lib/toolbar/toolbar.component.ts
```diff
@@ -183,4 +208,27 @@ export class ToolbarComponent implements AfterViewInit, OnDestroy {
...
+
+                    if (!buttons.length) {
```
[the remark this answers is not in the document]
DmytroPutrin: done

dmytrofirsov (approved):

dmytrofirsov (approved):

Vita-Karas (approved):

PR #330, opened 2021-06-11
Feature/NUI-6101 keyboard navigating in the wizard v2

Vita-Karas:


packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.less
```diff
@@ -31,6 +31,11 @@
         .nui-step-header-ripple {
             pointer-events: none;
         }
+
+        &:focus {
+            .focus-outline();
+            border-radius: 5px;
```
Blakesalot, line 37: Is there a chance that we should use `@nui-radius-default` found [here](https://github.com/solarwinds/nova/blob/8928c3359be235534175f58815665301be0d4838/packages/bits/src/styles/nui-framework-variables.less#L71)? It has a value of 3px. If it needs to be 5px according to UX, maybe it would be good to put a variable to use in this file:

https://github.com/solarwinds/nova/blob/aba6f99353d6ce384cf8748c7f1fdb5a2e12c2b4/packages/bits/src/lib/wizard-v2/wizard-vars.less

Vita-Karas: We can use @nui-radius-default here. This work in progress.

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.less
```diff
@@ -31,6 +31,11 @@
         .nui-step-header-ripple {
             pointer-events: none;
         }
+
+        &:focus {
+            .focus-outline();
```
Blakesalot, line 36: Same comment as this one: https://github.com/solarwinds/nova/pull/332/files#r651157657

DmytroPutrin (approved):

dmytrofirsov (dismissed):

packages/bits/src/lib/wizard-v2/wizard-horizontal/wizard-horizontal.component.html
```diff
@@ -1,5 +1,6 @@
 <div class="nui-wizard-horizontal-header-container">
-    <div class="nui-step-header-wrapper" *ngFor="let step of steps; let i = index; let isLast = last">
+    <div class="nui-step-header-wrapper" *ngFor="let step of steps; let i = index; let isFirst = first">
+        <div *ngIf="!isFirst" class="nui-wizard-horizontal-line" [ngClass]="{'nui-wizard-horizontal-line-completed': steps.get(i-1)?.completed && step.completed || selectedIndex === i}"></div>
```
Blakesalot, line 3: Hi, Vita. I think we should avoid adding a function call to the template (`steps.get()`) since function calls in templates create inefficiencies in change detection. I would be okay with using the existing getStepState function instead for now since we have [this](https://jira.solarwinds.com/browse/NUI-6129) ticket to come up with a better solution. That way, at least we know the function will be replaced with something more efficient soon.

packages/bits/demo/src/components/demo/wizard-v2/wizard-busy/wizard-busy.example.component.html
```diff
@@ -1,6 +1,6 @@
 <div nui-busy [busy]="busy" class="example-container">
     <div class="busy-content">
-        <nui-wizard-horizontal>
+        <nui-wizard-horizontal linear>
```
Blakesalot, line 3: What is this for exactly? Whatever it's doing I wonder if it could be built into the wizard instead of requiring the consumer to do it.

Blakesalot: Out of curiosity, if it's an input, why doesn't it use notation like this: `[linear]="true"`?

dmytrofirsov: Typically, if there is no input specified for the component, the input remains `undefied` (if it wasn't initialized in the first place). But in case the input is specified without a value assigned to it, the input receives an empty string by default. 
Angular Material team treats`""` as a truthy value for their input here with this function:

```
export function coerceBooleanProperty(value: any): boolean {  return value != null &amp;&amp; `${value}` !== 'false';}
```

image

Blakesalot: I think I would prefer a more explicit `[linear]="true"` to prevent confusion and also to avoid relying on Material's coersion which could be changed at any time. Additionally, can we set it to true by default in the wizard and have consumers turn it off if they want to override the behavior? It seems like linear would be the more common case.

dmytrofirsov: I think this is pretty neat and readable, and explicit enough thanks to their docs. I wouldn't be worrying about material changing things since we're relying on them so heavily if there will be any breaking changes not only this will blow up. Personally, I prefer following the official docs and guides because Material is responsible for the support of their docs and the functionality they describe. As consumers, we don't have to inspect their source code, usually reading the docs is enough. If we deviate from the prescribed instructions we'll be to blame due to the component misuse, even if there is a bug on their side. Just like the empty string may stop working in their method over some time due to the method changes, the directly passed in `true` boolean may stop working either because of the same reason. 

We may have it as a default option though,  I also see this being the most used option.

Blakesalot: Okay, if it's that way in their docs, that sounds fine to me. And, setting it to linear by default sounds good too. Maybe we could have a link to the Material usage from our docs as well?

DmytroPutrin (approved):

packages/bits/src/lib/wizard-v2/wizard-horizontal/wizard-horizontal.component.html
Vita-Karas: I created a `getStep` pipe for avoiding multiple function calls in template

packages/bits/src/lib/wizard-v2/wizard-horizontal/wizard-horizontal.component.ts
```diff
@@ -41,4 +41,10 @@ export class WizardHorizontalComponent extends WizardDirective {
 
     /** Whether the label should display in bottom or end position. */
     @Input() labelPosition: "top" | "end" = "top";
+    
+
+    ngOnInit(): void {
+        // we should be checked the validity of previous steps
```
dmytrofirsov, line 47:
```suggestion
        // Checking the validity of previous steps by default.
```

packages/bits/src/pipes/get-step.pipe.ts
```diff
@@ -0,0 +1,26 @@
...
+ *   steps | getStep:index
+ *
+ * __Example :__
+ *   <code>{{ steps | getStep:3 }}</code>
+ *
+ */
+@Pipe({
+    name: "getStep",
+})
+export class GetStep implements PipeTransform {
```
dmytrofirsov, line 22: This pipe is very specific to the wizard. Do we really want to export it to our consumers? Maybe, we may keep with the wizard

Vita-Karas: Moved, thanks

DmytroPutrin (approved):

dmytrofirsov (approved):

PR #332, opened 2021-06-11
made expander accessible

dmytrofirsov:


DmytroPutrin (approved):

Vita-Karas (approved):

packages/bits/src/lib/expander/expander.component.less
```diff
@@ -1,8 +1,15 @@
 @import (reference) "../../styles/mixins/expander-base";
+@import (reference) "../../styles/mixins/focus.less";
 
 .nui-expander {
     .expander();
 
+    &__header {
+        &:focus {
+            .focus-outline();
```
Blakesalot, line 9: I wonder if we should avoid using this mixin until the setting of the z-index is removed:

https://github.com/solarwinds/nova/blob/3bffdc91025981eee619e3bafb53aefb993fc05e/packages/bits/src/styles/mixins/focus.less#L3-L4

To be safe, maybe we should just duplicate those styles here without the z-index.

By the way, I'm not sure why the mixin has an outline rule with a value of `none`, but chrome reads that as an invalid value for outline. However, chrome does like `0` for that rule value. See this comment here for reference:

https://github.com/solarwinds/nova/pull/310#discussion_r641789686

dmytrofirsov: @Blakesalot `outline: none` looks pretty valid to me
image

ivankozlovskyi (approved):

DmytroPutrin (approved):

packages/bits/src/lib/expander/expander.component.less
Blakesalot: Okay, thanks for checking. I'm not sure what I saw then :)

Blakesalot: @dmytrofirsov, aside from that, are you considering whether or not to use the mixin since it sets a z-index?

dmytrofirsov: I decided to make changes with a comment in the code

packages/bits/src/lib/menu/menu-key-control.service.ts
```diff
@@ -146,12 +148,15 @@ export class MenuKeyControlService implements OnDestroy {
...
     private handleClosedKeyDown(event: KeyboardEvent): void {
+        event.stopPropagation();
+
         // prevent opening on enter and prevent scrolling page on key down/key up when focused
         if (this.shouldBePrevented(event)) {
             event.preventDefault();
         }
 
         if (event.code === KEYBOARD_CODE.ARROW_DOWN) {
+            event.preventDefault();
```
Vita-Karas, line 159: I think this call of `preventDefault()` is redundant because we call it in the first `if` statement

DmytroPutrin (approved):

Vita-Karas (approved):

PR #348, opened 2021-06-17
[NUI-5924] add keyboard navigation for timepicker

DmytroPutrin:


packages/bits/src/lib/time-picker/time-picker.component.ts
```diff
@@ -142,14 +152,34 @@ export class TimePickerComponent implements OnInit, OnDestroy, OnChanges, AfterV
...
 
     onTouched() {}
 
+    @HostListener("keydown", ["$event"])
+    public onKeyDown(event: KeyboardEvent): void {
+        this.keyboardService.onKeyDown(event);
+    }
+
+    @HostListener("focusout")
+    onFocusOut(): void {
```
dmytrofirsov, line 171: lets use access modifiers to be consistent across the code

packages/bits/src/lib/time-picker/time-picker.component.ts
```diff
@@ -142,14 +152,34 @@ export class TimePickerComponent implements OnInit, OnDestroy, OnChanges, AfterV
...
+        if (!this.overlay.showing || !document.activeElement) {
+            return;
+        }
+
+        setTimeout(() => {
```
dmytrofirsov, line 176: Why do we need `setTiemout` here? An explicit comment would be great to have for the future reference

packages/bits/src/lib/time-picker/time-picker.component.ts
```diff
@@ -142,14 +152,34 @@ export class TimePickerComponent implements OnInit, OnDestroy, OnChanges, AfterV
         this.overlay.clickOutside
             .pipe(takeUntil(this.onDestroy$))
             .subscribe(_ => this.overlay.hide());
-        
+
         this.initPopupUtilities();
+        this.keyboardService.initService(this.popup.menuItems, this.popup, this.overlay, this.menuTrigger.nativeElement);
```
dmytrofirsov, line 157: if we're passing in the `this.popup`, we can probably retrieve the `this.popup.menuItems` from it later, can't we?

packages/bits/src/lib/time-picker/time-picker.component.less
```diff
@@ -12,11 +12,21 @@
...
 
-    &__icon {
+    &__trigger {
         cursor: pointer;
         position: absolute;
         top: @nui-space-xs + @nui-space-xxs;
         right: @nui-space-xs + @nui-space-xxs;
+        outline: none;
+
+        &:focus-within .nui-timepicker__icon {
```
dmytrofirsov, line 22: you could have just a nested `:focus-within` inside the `&__icon` below.

I'm still concerned about the wrapper, lets think how we can go about it in a different way. 

Please check if the screen reader really reads the contents of the nested `svg`s. If not maybe the `aria-ignore` is not relevant for the icon host.

DmytroPutrin: thanks, reader doesn't go to svg, I removed wrapper.

dmytrognatyshyn (approved):

ivankozlovskyi (approved):

PR #352, opened 2021-06-22
Feature/n UI 6126 add tooltip to wizard header

DmytroPutrin:


packages/bits/demo/src/components/demo/wizard-v2/wizard-tooltip/wizard-tooltip-example.component.html
```diff
@@ -0,0 +1,41 @@
+<nui-wizard-horizontal #wizardComponent>
+
+    <!--  content of the 1st step -->
+    <nui-wizard-step-v2 i18n-label label="Very long header text. Very long header text">
+        <!-- content of the current step -->
+        <span i18n>Step with long header text</span>.
+
+        <!-- content of the footer -->
+        <ng-template wizardStepFooter>
```
Blakesalot, line 9: @DmytroPutrin, do you mind creating a ticket to change the selector for this directive to `nuiWizardStepFooter`? It's better to prefix our directives with `nui` to avoid the possibility of naming conflicts on the consumer side. Feel free to change it in the scope of this PR as an alternative to creating a ticket.

packages/bits/demo/src/components/demo/wizard-v2/wizard-tooltip/wizard-tooltip-example.component.html
```diff
@@ -0,0 +1,41 @@
...
+            <ng-container *ngTemplateOutlet="wizardNext"></ng-container>
+        </ng-template>
+    </nui-wizard-step-v2>
+
+    <!--  Step with custom label template -->
+    <nui-wizard-step-v2 i18n-label tooltip="Custom Label Template">
+        <ng-template wizardStepLabel>
```
Blakesalot, line 16: Similar to previous comment, do you mind creating a ticket for renaming this selector to `nuiWizardStepLabel`? Alternately, doing it in this PR is fine as well.

packages/bits/demo/src/components/demo/wizard-v2/wizard-tooltip/wizard-tooltip-example.component.html
```diff
@@ -0,0 +1,41 @@
...
+        </ng-template>
+        <span i18n>We finally reached our last step</span>!
+        <!-- Notice how both label and footer can be placed anywhere in the content of wizard step -->
+        <ng-template wizardStepFooter>
+            <ng-container *ngTemplateOutlet="wizardPrevious"></ng-container>
+        </ng-template>
+    </nui-wizard-step-v2>
+
+    <ng-template #wizardPrevious>
+        <button type="button" nuiWizardPrevious nui-button icon="caret-left">
```
Blakesalot, line 27: needs `i18n` attribute

packages/bits/demo/src/components/demo/wizard-v2/wizard-tooltip/wizard-tooltip-example.component.html
```diff
@@ -0,0 +1,41 @@
...
+        </button>
+    </ng-template>
+
+    <ng-template #wizardNext>
+        <button nuiWizardNext
+                type="button"
+                nui-button
+                displayStyle="primary"
+                icon="caret-right"
+                iconRight="true"
```
Blakesalot, line 38: needs `i18n` attribute

packages/bits/demo/src/components/demo/wizard-v2/wizard-tooltip/wizard-tooltip-example.component.html
```diff
@@ -0,0 +1,41 @@
...
+        >Next</button>
```
Blakesalot, line 39: Do you mind putting the closing `>` on line 38 with the last attribute for formatting consistency?

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.ts
```diff
@@ -128,4 +151,13 @@ export class WizardStepHeaderComponent extends CdkStepHeader implements AfterVie
             [STEP_STATE.ERROR]: this.wizardConfig.stepState?.error,
         }
     }
+
+    private detectLongLabel(): void {
+        if (this.labelEl) {
+            const el: HTMLElement = this.labelEl.nativeElement;
+
+            this.isLabelOverflow = el.scrollWidth > el.offsetWidth;
```
Blakesalot, line 159: Do we know for sure whether UX doesn't want us to show the label on hover regardless of the overflow status? It would be nice to not have to worry about making this calculation and just have the label available at all times regardless of the length.

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.ts
```diff
@@ -95,6 +105,7 @@ export class WizardStepHeaderComponent extends CdkStepHeader implements AfterVie
 
     ngAfterViewInit(): void {
         this._focusMonitor.monitor(this._elementRef, true);
+        this.detectLongLabel();
```
Blakesalot, line 108: Do we need handle dynamic changes in the available space and label length? If so, this would need to be done in ngOnChanges as well.

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.less
```diff
@@ -7,7 +7,6 @@
     &-header {
         overflow: hidden;
         outline: none;
-        cursor: pointer;
```
Blakesalot, line 10: Do we still want to show the pointer cursor for when the user wants to navigate directly to a step by clicking on the header?

packages/bits/demo/src/components/demo/wizard-v2/wizard-docs/wizard-docs.component.html
```diff
@@ -104,6 +104,17 @@ <h3>Wizard Step Change Event Example</h3>
     <nui-wizard-step-change-example></nui-wizard-step-change-example>
 </nui-example-wrapper>
 
+<h3>Wizard Tooltip</h3>
+<p>
+    You are able add tooltip to wizard. By default if label text overflow container we show tooltip.
+    Note that it doesn't work with label with custom template. In this case you should add tooltip explicitly. For example:
+    <br>
+</p>
+<pre>< nui-wizard-step-v2 tooltip="Custom Label Template" ></pre>
```
Blakesalot, line 113: I feel like we can leave out this code snippet, and just point consumers to the source code in the example wrapper like we usually do.

packages/bits/demo/src/components/demo/wizard-v2/wizard-docs/wizard-docs.component.html
```diff
@@ -104,6 +104,17 @@ <h3>Wizard Step Change Event Example</h3>
     <nui-wizard-step-change-example></nui-wizard-step-change-example>
 </nui-example-wrapper>
 
+<h3>Wizard Tooltip</h3>
+<p>
+    You are able add tooltip to wizard. By default if label text overflow container we show tooltip.
+    Note that it doesn't work with label with custom template. In this case you should add tooltip explicitly. For example:
```
Blakesalot, line 110: Maybe something like this?

```suggestion
    Note that if you're using a custom template for the step label you need to specify the tooltip text explicitly. For how to do this, take a look at the template source code of the example below.
```

packages/bits/demo/src/components/demo/wizard-v2/wizard-tooltip/wizard-tooltip-example.component.html
DmytroPutrin: done

packages/bits/demo/src/components/demo/wizard-v2/wizard-tooltip/wizard-tooltip-example.component.html
DmytroPutrin: done

packages/bits/demo/src/components/demo/wizard-v2/wizard-docs/wizard-docs.component.html
DmytroPutrin: done, thanks

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.ts
DmytroPutrin: I guess we don't need to handle width dynamically. It's really edge case if user starts resize browser window. And dom calculation for this could cost performance. Other case will be cover by NUI-5068.

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.ts
DmytroPutrin: User could pass tooltip as input and he will be able see tooltip regardless of the length. This calculation for case when user doesn't specify tooltip and label text overflows container.

Blakesalot: Okay, in that case, we may want to ask UX whether we should show the tooltip at all times regardless of whether the `tooltip` input is used. The current behavior of showing a tooltip at all times on hover when the `tooltip` input is used but only showing it when the label text overflows when the tooltip input isn't being used may be seen as inconsistent from the consumer and/or end-user point of view.

This brings to mind an additional question: Maybe the consumer-provided template itself should be responsible for showing the tooltip (instead of exposing a separate `tooltip` input. The purpose of template injection is to allow consumers to put whatever they want in a certain spot on the DOM. So, in this case if they want a tooltip on their template, they can just inject a nui-tooltip along with the label contents. That way, the label template and its associated tooltip are all in one spot instead of having a special input for it. I think I like that better. What do you think?

packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.ts
Blakesalot: Actually, I don't think we should consider resizing to be an edge case. If we have logic to show a tooltip on overflow, I think it's important that the logic works without having to refresh the page.

Blakesalot: On the other hand, if UX agrees that we can unconditionally show the tooltip regardless of overflow status, this problem would go away.

packages/bits/demo/src/components/demo/wizard-v2/wizard-tooltip/wizard-tooltip-example.component.html
```diff
@@ -0,0 +1,45 @@
...
+    </nui-wizard-step-v2>
+
+    <!--  Step with custom label template -->
+    <nui-wizard-step-v2>
+        <ng-template nuiWizardStepLabel >
+            <span i18n-nuiTooltip nuiTooltip="Custom Label template">
+                <i>Custom Label template</i>
+            </span>
+        </ng-template>
+        <span i18n>We finally reached our last step</span>!
```
Vita-Karas, line 21: exclamation mark after the closed `span`

Vita-Karas (dismissed):

packages/bits/demo/src/components/demo/wizard-v2/wizard.module.ts
```diff
@@ -22,7 +22,7 @@ import {
     NuiSelectV2Module,
     WIZARD_CONFIG,
     IWizardConfig,
-    WIZARD_CONFIG_DEFAULT,
+    WIZARD_CONFIG_DEFAULT, NuiTooltipModule,
```
dmytrofirsov, line 25: each added import should start from a new line

ivankozlovskyi (approved):

dmytrognatyshyn (approved):

PR #354, opened 2021-06-23
Feature/nui 5977 legend metadata series

baggiosw:


packages/charts/src/renderers/xy-renderer.ts
```diff
@@ -18,5 +28,18 @@ export abstract class XYRenderer<TA extends IAccessors> extends Renderer<TA> {
             y: scales.y.convert(dataSeries.accessors.data.y(point, index, dataSeries.data, dataSeries)),
         };
     }
+    public getDataPointIndex(series: IDataSeries<IXYAccessors>, values: { [p: string]: any }, scales: Scales): number {
+        if (isUndefined(values.x)) {
+            return DATA_POINT_INTERACTION_RESET;
+        }
 
+        // @ts-ignore
```
baggiosw, line 36: Not sure why I needed this ts-ignore but it works with it I would like a suggestion on how to fix this if possible

packages/charts/examples/components/demo/advanced-usage/legend/legend-metadata/legend-metadata-example.component.html
```diff
@@ -0,0 +1,37 @@
+<div class="nui-chart-layout">
+    <div class="chart">
+        <nui-chart class="w-100" [chart]="chart"></nui-chart>
+    </div>
+    <div class="legend">
+        <nui-legend seriesUnitLabel="Mbps" i18n-seriesUnitLabel
+                    [active]="chartAssist.isLegendActive"
+                    [interactive]="true"
+                    (mouseleave)="chartAssist.resetVisibleSeries()">
+            <nui-legend-series [interactive]="false"
```
Blakesalot, line 10: A comment above this series might be good. Something like:

```
<!-- This is the metadata series -->
```

packages/charts/examples/components/demo/advanced-usage/legend/legend-metadata/legend-metadata-example.component.html
```diff
@@ -0,0 +1,37 @@
...
+                               [descriptionPrimary]="avgSeries['name']">
+                <nui-rich-legend-tile [value]="chartAssist.getHighlightedValue(avgSeries, 'y', undefined, 'y1')"
+                                      backgroundColor="black"
+                                      color="white">
+                </nui-rich-legend-tile>
+            </nui-legend-series>
+            <nui-legend-series *ngFor="let legendSeries of chartAssist.legendSeriesSet; trackBy: chartAssist.seriesTrackByFn"
```
Blakesalot, line 17: A comment above this element might be good. Something like:

```
<!-- These are the series displayed on the chart -->
```

packages/charts/examples/components/demo/advanced-usage/legend/legend-metadata/legend-metadata-example.component.html
```diff
@@ -0,0 +1,37 @@
...
+    <div class="legend">
+        <nui-legend seriesUnitLabel="Mbps" i18n-seriesUnitLabel
+                    [active]="chartAssist.isLegendActive"
+                    [interactive]="true"
+                    (mouseleave)="chartAssist.resetVisibleSeries()">
+            <nui-legend-series [interactive]="false"
+                               [descriptionPrimary]="avgSeries['name']">
+                <nui-rich-legend-tile [value]="chartAssist.getHighlightedValue(avgSeries, 'y', undefined, 'y1')"
+                                      backgroundColor="black"
+                                      color="white">
```
Blakesalot, line 14: Can we use nova colors here somehow?

packages/charts/examples/components/demo/advanced-usage/legend/legend-metadata/legend-metadata-example.component.html
```diff
@@ -0,0 +1,37 @@
...
+            </nui-legend-series>
+            <nui-legend-series *ngFor="let legendSeries of chartAssist.legendSeriesSet; trackBy: chartAssist.seriesTrackByFn"
+                               [descriptionPrimary]="legendSeries['name']"
+                               (mouseenter)="chartAssist.emphasizeSeries(legendSeries.id)"
+                               [seriesRenderState]="chartAssist.renderStatesIndex[legendSeries.id]?.state"
+                               [isSelected]="!chartAssist.isSeriesHidden(legendSeries.id)"
+                               (isSelectedChange)="chartAssist.toggleSeries(legendSeries.id, $event)">
+
+                <!-- Specify 'undefined' for the formatter key since we're not applying any special formatting to the data values shown in the legend -->
+                <!-- Specify the 'y1' data accessor to let the chart assist know how to access the data for the legend. -->
```
Blakesalot, line 25: Maybe these comments could be moved to just above line 12 where the usage is encountered on the page for the first time.

packages/charts/examples/components/demo/advanced-usage/legend/legend-metadata/legend-metadata-example.component.ts
```diff
@@ -0,0 +1,156 @@
...
+    selector: "area-chart-stack-example",
+    templateUrl: "./legend-metadata-example.component.html",
+})
+export class LegendMetadataExampleComponent implements OnInit {
+    public chart: Chart;
+    public chartAssist: ChartAssist;
+    public avgSeries: IChartAssistSeries<ILineAccessors>
+
+    public ngOnInit() {
+        // areaGrid returns an XYGrid configured for displaying an area chart's axes and other grid elements.
```
Blakesalot, line 32: I think all the comments about area chart usage can be removed since we're focusing on the legend here.

packages/charts/examples/components/demo/advanced-usage/legend/legend-metadata/legend-metadata-example.component.ts
```diff
@@ -0,0 +1,156 @@
...
+
+        const averageData = getAverage(getData());
+        // We are using a different renderer so the metadata does not effect the domain
+        const noopRenderer = new XYRenderer({ignoreForDomainCalculation: true});
+        // Here we create an accessor for our average metadata
+        const avgAccessors = new XYAccessors();
+        // This is so the legend knows the value for the y
+        avgAccessors.data.y = (d) => d.value;
+        // Same as above we are setting the numeric value we want to visualize
+        avgAccessors.data.y1 = avgAccessors.data.y;
```
Blakesalot, line 61: I think these accessor modifications can be removed if you change the template usage of getHighlightedDataPoint from this:

```
chartAssist.getHighlightedValue(avgSeries, 'y', undefined, 'y1')
```
to this:
```
chartAssist.getHighlightedValue(avgSeries, 'y')"
```

packages/charts/examples/components/demo/advanced-usage/legend/legend-metadata/legend-metadata-example.component.ts
```diff
@@ -0,0 +1,156 @@
...
+        // Setting the color of the series to transparent so it doesn't render
+        avgAccessors.series.color = () => "var(--nui-color-bg-transparent)";
```
Blakesalot, line 63: Maybe we could change it to this and use it in the template since we no longer draw the series on the chart:

```suggestion
        avgAccessors.series.color = () => "var(--nui-color-bg-mid_dark)";
```

Blakesalot: Actually, maybe modifying the accessor is unnecessary and the color can just be used directly in the template.

packages/charts/examples/components/demo/advanced-usage/legend/legend-metadata/legend-metadata-example.component.ts
```diff
@@ -0,0 +1,156 @@
...
+        // Deleting the markers so it doesn't place a marker on the chart
+        delete avgAccessors.series.marker;
```
Blakesalot, line 65: Is this still necessary now that we're not drawing the series?

packages/charts/examples/components/demo/advanced-usage/legend/legend-metadata/legend-metadata-example.component.ts
```diff
@@ -0,0 +1,156 @@
...
+        const renderer = new AreaRenderer();
+
+        // In case of a area chart, the scale definitions are flexible.
+        // This example demonstrates a scenario with time on the X scale and a numeric value on the Y scale.
+        const scales: IXYScales = {
+            x: new TimeScale(),
+            y: new LinearScale(),
+        };
+
+        const averageData = getAverage(getData());
```
Blakesalot, line 53: Maybe `calculateAverageSeries`?

packages/charts/examples/components/demo/advanced-usage/legend/legend-metadata/legend-metadata-example.component.ts
```diff
@@ -0,0 +1,156 @@
...
+        // We are using a different renderer so the metadata does not effect the domain
```
Blakesalot, line 54: Maybe this:

```suggestion
        // We are using the base XYRenderer so the metadata does not get displayed on the chart
```

packages/charts/examples/components/demo/advanced-usage/legend/legend-metadata/legend-metadata-example.component.ts
```diff
@@ -0,0 +1,156 @@
...
+        const noopRenderer = new XYRenderer({ignoreForDomainCalculation: true});
```
Blakesalot, line 55: Maybe this?
```suggestion
        const metaDataRenderer = new XYRenderer({ignoreForDomainCalculation: true});
```

packages/charts/examples/components/demo/advanced-usage/legend/legend-metadata/legend-metadata-example.component.ts
baggiosw: I think it doesn't hurt so I would like to keep them if that's fine.

packages/charts/examples/components/demo/advanced-usage/legend/legend-metadata/legend-metadata-example.component.ts
```diff
@@ -0,0 +1,156 @@
...
+        // Deleting the markers so it doesn't place a marker on the chart
+        delete avgAccessors.series.marker;
+
+        this.avgSeries = {
+            ...averageData,
+            accessors: avgAccessors,
+            renderer: noopRenderer,
+            scales: scales,
+            showInLegend: false,
+            preprocess: false,
```
Blakesalot, line 73: It would be good to add a comment about why these need to be false.

packages/charts/examples/components/demo/advanced-usage/legend/legend-metadata/legend-metadata-example.component.ts
```diff
@@ -0,0 +1,156 @@
...
+    }
+    return {
+        id: "average",
+        name: "Average Speed",
+        data: arrAverage,
+    };
+}
+
+/* Chart data */
+function getData() {
```
Blakesalot, line 106: Maybe a return type:

```suggestion
function getData(): Partial<IDataSeries<XYAccessors>>[] {
```

packages/charts/src/renderers/accessors/xy-accessors.ts
```diff
@@ -8,11 +10,35 @@ export interface IXYDataAccessors {
...
+
+    constructor(public colorProvider = defaultColorProvider(), public markerProvider = defaultMarkerProvider()) {
+
+        this.data.defined = (d: any, i) => d.hasOwnProperty("defined") ? d.defined : true;
+
+        this.series = {
+            color: this.colorProvider ? this.colorProvider.get : undefined,
+            marker: this.markerProvider ? this.markerProvider.get : undefined,
+        };
+    }
```
Blakesalot, line 43: I'm wondering if we need the color and marker accessors at all since we're not drawing anything on the chart.

packages/charts/src/renderers/line/line-accessors.ts
```diff
@@ -37,14 +37,7 @@ export class LineAccessors extends XYAccessors implements ILineAccessors {
...
 
-    constructor(private colorProvider = defaultColorProvider(), private markerProvider = defaultMarkerProvider()) {
-        super();
-
-        this.data.defined = (d: any, i) => d.hasOwnProperty("defined") ? d.defined : true;
-
-        this.series = {
-            color: this.colorProvider ? this.colorProvider.get : undefined,
-            marker: this.markerProvider ? this.markerProvider.get : undefined,
-        };
```
Blakesalot, line 48: It makes more sense to me to leave these here instead of moving them up to the XYRenderer which doesn't draw anything on the chart.

packages/charts/src/renderers/xy-renderer.ts
```diff
@@ -1,8 +1,19 @@
...
+import { DATA_POINT_INTERACTION_RESET } from "../constants";
+import { UtilityService } from "../core/common/utility.service";
+import { IXYAccessors } from "./accessors/xy-accessors";
 
-export abstract class XYRenderer<TA extends IAccessors> extends Renderer<TA> {
+
+export class XYRenderer<TA extends IXYAccessors> extends Renderer<TA> {
+
+    public draw(renderSeries: IRenderSeries<TA>, rendererSubject: Subject<IRendererEventPayload>): void {
+
```
Blakesalot, line 15: Maybe a comment about why this is empty:

```suggestion
// This is empty to allow for series that represent metadata that may be shown in the legend but not visualized on the chart.
```

packages/charts/src/renderers/xy-renderer.ts
```diff
@@ -18,5 +29,17 @@ export abstract class XYRenderer<TA extends IAccessors> extends Renderer<TA> {
             y: scales.y.convert(dataSeries.accessors.data.y(point, index, dataSeries.data, dataSeries)),
         };
     }
+    public getDataPointIndex(series: IDataSeries<TA>, values: { [p: string]: any }, scales: Scales): number {
```
Blakesalot, line 32: newline would be good
```suggestion

    public getDataPointIndex(series: IDataSeries<TA>, values: { [p: string]: any }, scales: Scales): number {
```

Blakesalot: It looks like AreaRenderer has an identical implementation of this method, so I think AreaRenderer's override of getDataPointIndex can be removed.

packages/charts/src/renderers/xy-renderer.ts
```diff
@@ -18,5 +29,17 @@ export abstract class XYRenderer<TA extends IAccessors> extends Renderer<TA> {
...
+        if (isUndefined(values.x)) {
```
Blakesalot, line 33: And, I think the LineRenderer version of this method can be removed too with one minor addition here:

```suggestion
        if (!this.config.interactive || isUndefined(values.x)) {
```

packages/charts/examples/components/demo/advanced-usage/legend/legend-metadata/legend-metadata-example.component.html
```diff
@@ -0,0 +1,38 @@
...
+    <div class="legend">
+        <nui-legend seriesUnitLabel="Mbps" i18n-seriesUnitLabel
+                    [active]="chartAssist.isLegendActive"
+                    [interactive]="true"
+                    (mouseleave)="chartAssist.resetVisibleSeries()">
+            <!-- This is the metadata series -->
+            <nui-legend-series [interactive]="false"
+                               [descriptionPrimary]="avgSeries['name']">
+                <!-- Specify 'undefined' for the formatter key since we're not applying any special formatting to the data values shown in the legend -->
+                <!-- Specify the 'y1' data accessor to let the chart assist know how to access the data for the legend. -->
```
Blakesalot, line 14: Now that we've simplified the usage of getHighlightedValue here, it seems like we should move the comments back to line 27.

packages/charts/examples/components/demo/advanced-usage/legend/legend-metadata/legend-metadata-example.component.ts
```diff
@@ -0,0 +1,157 @@
...
+        }));
+        // Combining both sets in an array of series
+        seriesSet = [...seriesSet, this.avgSeries];
+
+        // Finally, pass the series set to the chart's update method
+        this.chartAssist.update(seriesSet as IChartSeries<IAreaAccessors>[]);
+    }
+}
+
+function calculateAverageSeries(dataArr: Partial<IDataSeries<XYAccessors>>[]): Partial<IDataSeries<XYAccessors>> {
```
Blakesalot, line 84: We've been using `seriesSet` as the symbol name when we're referring to multiple series. Do you mind changing it:

```suggestion
function calculateAverageSeries(seriesSet: Partial<IDataSeries<XYAccessors>>[]): Partial<IDataSeries<XYAccessors>> {
```

packages/charts/examples/components/demo/advanced-usage/legend/legend-metadata/legend-metadata-example.component.ts
```diff
@@ -0,0 +1,157 @@
...
+        this.avgSeries = {
+            ...averageData,
+            accessors: avgAccessors,
+            renderer: metaDataRenderer,
+            scales: scales,
+            // showInLegend is false because we manually add our own series
+            showInLegend: false,
+        }
+        // Here we assemble the complete chart series.
+        let seriesSet: Partial<IChartSeries<IAccessors>>[] = getData().map(d => ({
```
Blakesalot, line 70: I just noticed that we're invoking getData twice. We could optimize that by putting the result in a variable and reusing it here.

packages/charts/examples/components/demo/advanced-usage/legend/legend-metadata/legend-metadata-example.component.ts
```diff
@@ -0,0 +1,157 @@
...
+        // This is so the legend knows the value for the y
+        avgAccessors.data.y = (d) => d.value;
+
+        this.avgSeries = {
+            ...averageData,
+            accessors: avgAccessors,
+            renderer: metaDataRenderer,
+            scales: scales,
+            // showInLegend is false because we manually add our own series
+            showInLegend: false,
```
Blakesalot, line 67: Don't we need preprocess to be false here as well?

packages/charts/src/renderers/accessors/xy-accessors.ts
```diff
@@ -8,11 +11,26 @@ export interface IXYDataAccessors {
...
+    /** Marker for the series */
+    marker?: SeriesAccessor;
+    /** Additional custom keys to match the base interface */
+    [key: string]: SeriesAccessor | undefined;
+}
+export interface IXYAccessors extends IAccessors {
+    data: IXYDataAccessors;
+    /** Series level accessors - e.g. for colors, markers, etc. */
+    series: IXYSeriesAccessors;
+}
```
Blakesalot, line 26: Do we still need these interfaces? It seems like we could just remove the abstract keyword from the original XYAccessors class and that would be all that's needed.

packages/charts/src/renderers/line/line-accessors.ts
```diff
@@ -37,7 +37,7 @@ export class LineAccessors extends XYAccessors implements ILineAccessors {
     public data: ILineDataAccessors;
     public series: ILineSeriesAccessors;
 
-    constructor(private colorProvider = defaultColorProvider(), private markerProvider = defaultMarkerProvider()) {
+    constructor(public colorProvider = defaultColorProvider(), public markerProvider = defaultMarkerProvider()) {
```
Blakesalot, line 40: is changing them to public needed for some reason? doesn't hurt I guess.

packages/charts/src/renderers/area/area-renderer.ts
```diff
@@ -103,20 +103,6 @@ export class AreaRenderer extends XYRenderer<IAreaAccessors> {
         path.attr("d", areaFunc(validatedData) ?? "");
     }
 
-    /** See {@link Renderer#getDataPointIndex} */
```
Blakesalot, line 106: I think we can remove the getDataPointIndex override from the LineRenderer as well.

packages/charts/examples/components/demo/advanced-usage/legend/legend-metadata/legend-metadata-example.component.ts
```diff
@@ -0,0 +1,157 @@
...
+function calculateAverageSeries(dataArr: Partial<IDataSeries<XYAccessors>>[]): Partial<IDataSeries<XYAccessors>> {
+    let arrAverage = [];
+    const dataLength =  dataArr[0].data?.length ?? 0;
+    const numOfSeries = dataArr.length
+    if(!dataArr.length) {
+        return {
+            id: "average",
+            name: "Average Speed",
+            data: [],
+        }
```
Blakesalot, line 93: I guess this default return value could be stored in a variable above this. Then, the ultimate return value at line 104 could be mutated like this:

```
return { ...averageSeries, data: arrAverage };
```

That way you wouldn't have to repeat the id and name strings.

packages/charts/examples/components/demo/advanced-usage/legend/legend-metadata/legend-metadata-example.component.ts
```diff
@@ -0,0 +1,157 @@
...
+            avg += series?.data?.[n].value ?? 0;
+        }
+        avg = avg/numOfSeries
+        arrAverage.push({ x: dataArr[0].data?.[n].x, value: avg });
+    }
+    return {
+        id: "average",
+        name: "Average Speed",
+        data: arrAverage,
+    };
```
Blakesalot, line 108: Suggestion from above. This could be:

```
return { ...averageSeries, data: arrAverage };
```

packages/charts/examples/components/demo/advanced-usage/legend/legend-metadata/legend-metadata-example.component.ts
baggiosw: Nope since we aren't drawing it

packages/charts/src/renderers/line/line-accessors.ts
baggiosw: No reason just doing it for ease

packages/charts/examples/components/demo/advanced-usage/legend/legend-metadata/legend-metadata-example.component.ts
Blakesalot: But, I think the stackedArea series pre-processor will still process it. So, if the avgSeries were the first series in the set, I think it would get stacked on the bottom and the series that do get drawn would appear to be floating on top of it.

packages/charts/src/renderers/xy-renderer.ts
Blakesalot: We should transfer this comment from LineRenderer;

```suggestion
    /** See {@link Renderer#getDataPointIndex} */
    public getDataPointIndex(series: IDataSeries<TA>, values: { [p: string]: any }, scales: Scales): number {
```

packages/charts/src/renderers/xy-renderer.ts
```diff
@@ -1,8 +1,15 @@
...
+import isUndefined from "lodash/isUndefined";
+import { DATA_POINT_INTERACTION_RESET } from "../constants";
+import { UtilityService } from "../core/common/utility.service";
 
-export abstract class XYRenderer<TA extends IAccessors> extends Renderer<TA> {
+export class XYRenderer<TA extends IAccessors> extends Renderer<TA> {
+    // This is empty to allow this renderer to be used for series that represent metadata that may be shown in the legend but not visualized on the chart.
+    public draw(renderSeries: IRenderSeries<TA>, rendererSubject: Subject<IRendererEventPayload>): void {}
 
     public getDataPointPosition(dataSeries: IDataSeries<TA>, index: number, scales: Scales): IPosition | undefined {
```
Blakesalot, line 14: It would be good to add this comment here:
```suggestion
    /** See {@link Renderer#getDataPointPosition} */
    public getDataPointPosition(dataSeries: IDataSeries<TA>, index: number, scales: Scales): IPosition | undefined {
```

packages/charts/src/renderers/xy-renderer.ts
```diff
@@ -1,8 +1,15 @@
...
+import { IRenderSeries } from "./types";
+import { Subject } from "rxjs";
+import isUndefined from "lodash/isUndefined";
+import { DATA_POINT_INTERACTION_RESET } from "../constants";
+import { UtilityService } from "../core/common/utility.service";
 
-export abstract class XYRenderer<TA extends IAccessors> extends Renderer<TA> {
+export class XYRenderer<TA extends IAccessors> extends Renderer<TA> {
+    // This is empty to allow this renderer to be used for series that represent metadata that may be shown in the legend but not visualized on the chart.
+    public draw(renderSeries: IRenderSeries<TA>, rendererSubject: Subject<IRendererEventPayload>): void {}
```
Blakesalot, line 12: It would be good to add this comment here:

```suggestion
    /** See {@link Renderer#draw} */
    public draw(renderSeries: IRenderSeries<TA>, rendererSubject: Subject<IRendererEventPayload>): void {}
```

Blakesalot (dismissed):

Blakesalot (approved):

elenates (approved):

PR #364, opened 2021-06-28
NUI-6034:Nui table in Firefox: column sizes are ignored, column widths change on scroll

madelyneh:


packages/bits/spec/components/table/table.visual.ts
```diff
@@ -92,6 +92,11 @@ describe(`Visual tests: ${name}`, () => {
 
         await expanders.virtualScrollStickyHeader.click();
         await camera.say.cheese("Virtual Scroll Table with Sticky Header");
+
+        await expanders.virtualScrollStickyHeader.click();
```
Blakesalot, line 96: doesn't this close the expander since we already opened it at line 93?

packages/bits/src/lib/table/table-cell/table-cell.directive.ts
```diff
@@ -81,6 +83,16 @@ export class TableCellDirective extends CdkCell implements OnInit, OnDestroy, On
                     this.rightEdgeActive = !this.rightEdgeActive;
                 });
         }
+
+        this.tableStateHandlerService.columnWidthSubject.subscribe(() => {
+            const columnWidth = this.tableStateHandlerService.getColumnWidth(this.columnDef.name);
+
+            if (columnWidth > 45) {
+                if (this.tableStateHandlerService.getColumnWidthFixed(this.columnDef.name)) {
+                    elem.style.width = elem.style.maxWidth = elem.style.minWidth = columnWidth + "px";
```
Blakesalot, line 92: I think it would be good to put these assignments on separate lines. One assignment per line is less error prone as the person reading the code might miss an assignment in the middle of a line.

packages/bits/src/lib/table/table-cell/table-cell.directive.ts
```diff
@@ -3,6 +3,7 @@ import { ChangeDetectorRef, Directive, ElementRef, HostBinding, HostListener, In
 import _get from "lodash/get";
```
Blakesalot, line 3: I'm wondering if the changes in this file are necessary. Maybe they are necessary, but I would think we could just set the header cells and the body cells would follow suit automatically. Do you mind checking?

packages/bits/src/lib/table/table-cell/table-cell.directive.ts
```diff
@@ -3,6 +3,7 @@ import { ChangeDetectorRef, Directive, ElementRef, HostBinding, HostListener, In
...
 import { Subscription } from "rxjs";
 import { filter } from "rxjs/operators";
+import {FIXED_WIDTH_CLASS} from "../constants";
```
Blakesalot, line 6: It doesn't look like this file is using this constant. Should we be using this constant as a filter below?

packages/bits/src/lib/table/table-cell/table-column-def.directive.ts
```diff
@@ -30,18 +30,15 @@ export class TableColumnDefDirective extends CdkColumnDef implements OnInit, OnC
...
         if (this.type === "icon") {
             this.tableStateHandlerService.setAlignment(this.name, "align-center");
         }
     }
 
     ngOnChanges(changes: SimpleChanges) {
-        if (changes.columnWidth && !changes.columnWidth.firstChange) {
-            this.tableStateHandlerService.setColumnWidth(this.name, changes.columnWidth.currentValue);
+        if (changes.columnWidth) {
+            this.tableStateHandlerService.setColumnWidth(this.name, changes.columnWidth.currentValue, (this.columnWidth ? true : false));
```
Blakesalot, line 40: I think this could be shortened to `!!this.columnWidth`

packages/bits/src/lib/table/table-cell/table-header-cell.component.ts
```diff
@@ -217,7 +220,13 @@ export class TableHeaderCellComponent extends CdkHeaderCell implements OnInit, O
         this.subscriptions.push(this.tableStateHandlerService.columnWidthSubject.subscribe(() => {
             const columnWidth = this.tableStateHandlerService.getColumnWidth(this.columnDef.name);
             if (columnWidth > 45) {
-                this.elementRef.nativeElement.style.width = columnWidth + "px";
+                if (this.tableStateHandlerService.getColumnWidthFixed(this.columnDef.name)) {
+                    this.fixedWIdth = true;
+                    elem.style.maxWidth = elem.style.minWidth = columnWidth + "px";
```
Blakesalot, line 225: it would be good to put these assignments on separate lines for clarity

packages/bits/src/lib/table/table-cell/table-header-cell.component.ts
```diff
@@ -217,7 +220,13 @@ export class TableHeaderCellComponent extends CdkHeaderCell implements OnInit, O
...
+                }
+            } else {
+                this.fixedWIdth = false;
+                elem.style.maxWidth = elem.style.minWidth= "";
```
Blakesalot, line 229: separate lines for clarity

packages/bits/src/lib/table/table-cell/table-header-cell.component.ts
```diff
@@ -85,6 +85,8 @@ export class TableHeaderCellComponent extends CdkHeaderCell implements OnInit, O
     get isColumnSorted(): boolean {
         return this.sortingState.isColumnSorted;
     }
+    @HostBinding("class.nui-fixed-width")
+    public fixedWIdth: boolean;
```
Blakesalot, line 89:
```suggestion
    public fixedWidth: boolean;
```

packages/bits/src/lib/table/table-cell/table-header-cell.component.ts
```diff
@@ -85,6 +85,8 @@ export class TableHeaderCellComponent extends CdkHeaderCell implements OnInit, O
     get isColumnSorted(): boolean {
         return this.sortingState.isColumnSorted;
     }
+    @HostBinding("class.nui-fixed-width")
```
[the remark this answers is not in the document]
Blakesalot: newline needed here too

packages/bits/src/lib/table/table-virtual-scroll/table-sticky-header.directive.ts
```diff
@@ -179,8 +179,11 @@ export class TableStickyHeaderDirective implements AfterViewInit, OnDestroy {
             const fixedWidth = headColumns[index].classList.contains(FIXED_WIDTH_CLASS);
             if (!fixedWidth) {
                 // Note: Assigning data cell width to the corresponding header column
-                // (using the style width if specified; otherwise, falling back to the offsetWidth)
-                headColumns[index].style.width = cell.style.width || `${cell.offsetWidth}px`;
+                // (using the style width if specified; otherwise, falling back to the clientWidth)
+                const cellWidth = (parseInt(cell.style.width) || cell.clientWidth) -1 + "px";
```
Blakesalot, line 183: do we actually need to change this to clientWidth? if not, we should probably stick with offsetWidth to avoid any surprise bugs with clientWidth.

packages/bits/src/lib/table/table-virtual-scroll/table-sticky-header.directive.ts
```diff
@@ -179,8 +179,11 @@ export class TableStickyHeaderDirective implements AfterViewInit, OnDestroy {
...
+                const headStyle = headColumns[index].style;
+
+                headStyle.width =  headStyle.minWidth = headStyle.maxWidth = cellWidth;
```
Blakesalot, line 186: separate lines for assignment is preferable

packages/bits/src/lib/table/table-virtual-scroll/table-sticky-header.directive.ts
```diff
@@ -281,7 +284,7 @@ export class TableStickyHeaderDirective implements AfterViewInit, OnDestroy {
         const theadPlaceholder: Node = this.headRef.cloneNode(true);
 
         // Note: making header invisible
-        this.renderer.setStyle(theadPlaceholder, "visibility", "collapse");
+        this.renderer.setStyle(theadPlaceholder, "display", "none");
```
Blakesalot, line 287: I can't remember. Was this needed?

packages/bits/src/lib/table/table.component.less
```diff
@@ -124,8 +124,7 @@
         }
 
         &.virtual-sticky {
-            transition-property: width;
-            transition-duration: 0.2s;
+            transition: max-width 0.2s ease,min-width 0.2s ease, width 0.2s ease;
```
Blakesalot, line 127:
```suggestion
            transition: max-width 0.2s ease, min-width 0.2s ease, width 0.2s ease;
```

packages/bits/src/lib/table/table-cell/table-cell.directive.ts
madelyneh: If we dont keep this then the header and body column do not stay in sync.

packages/bits/src/lib/table/table-virtual-scroll/table-sticky-header.directive.ts
madelyneh: yes this made the scrolling smoother.

packages/bits/src/lib/table/table-cell/table-cell.directive.ts
Blakesalot: okay, thanks

packages/bits/demo/src/components/demo/table/table-docs/table-docs.example.component.html
```diff
@@ -291,6 +291,9 @@ <h3>Virtual Scroll with Sticky Header</h3>
     <code>tableStickyHeader</code> directive supports runtime state switch by just binding a boolean value to it <code>[tableStickyHeader]="false"</code>
     or taking the <code>TableStickyHeaderDirective</code> reference via <code>ViewChild</code> decorator and updating the <code>tableStickyHeader</code> property.
 </nui-message>
+<nui-message [allowDismiss]="false" type="info">
+    To have columns with fixed width set the <code>columnWidth</code> property with the desired width.
```
Blakesalot, line 295: For even better clarity maybe this?

```suggestion
    To have columns with fixed width set the <code>columnWidth</code> input on the <code>nuiColumnDef</code> directive to the desired width in pixels.
```

packages/bits/src/lib/table/table-cell/table-header-cell.component.ts
```diff
@@ -86,6 +86,9 @@ export class TableHeaderCellComponent extends CdkHeaderCell implements OnInit, O
         return this.sortingState.isColumnSorted;
     }
 
+    @HostBinding(`class.${FIXED_WIDTH_CLASS}`)
```
Blakesalot, line 89: I just noticed we already have a binding for this at line 64. In that case, `widthFixed` can be just a simple property, and the condition in the existing binding can be changed to this:

```
return this.isIconCell || this.fixedWidth;
```

packages/bits/src/lib/table/table-cell/table-header-cell.component.ts
```diff
@@ -217,7 +221,15 @@ export class TableHeaderCellComponent extends CdkHeaderCell implements OnInit, O
         this.subscriptions.push(this.tableStateHandlerService.columnWidthSubject.subscribe(() => {
             const columnWidth = this.tableStateHandlerService.getColumnWidth(this.columnDef.name);
             if (columnWidth > 45) {
-                this.elementRef.nativeElement.style.width = columnWidth + "px";
+                if (this.tableStateHandlerService.getColumnWidthFixed(this.columnDef.name)) {
```
Blakesalot, line 224: It looks like we're missing the original code from before:

```suggestion
                elem.style.width = columnWidth + "px";
                
                if (this.tableStateHandlerService.getColumnWidthFixed(this.columnDef.name)) {
```

packages/bits/src/lib/table/table-cell/table-header-cell.component.ts
madelyneh: I cant just use the get one because It also needs to be set. see how i changed it.

Blakesalot (approved):

elenates (approved):

PR #366, opened 2021-06-29
N UI 5070 add ut for wizard v2

DmytroPutrin:


packages/bits/src/lib/wizard-v2/wizard-step-header/wizard-step-header.component.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+            it("should call createStepStateConfigMap method", () => {
+                const spy = spyOn(component, "createStepStateConfigMap" as never);
+
+                component.ngOnChanges(config);
+                expect(spy).toHaveBeenCalled();
+            });
+        });
+
+        describe("stringLabel", () => {
+            afterEach(() => {
```
lukasholcik, line 74: Why are these necessary? Each unit test runs a separate instance of the component, so having this sort of cleanup shouldn't be required.

DmytroPutrin: you are right, thnaks

dmytrofirsov (approved):

Vita-Karas (approved):

dmytrognatyshyn (approved):

PR #369, opened 2021-06-30
hotfix: remove disabled class for step header

DmytroPutrin:


Vita-Karas:
Remove the style for the disabled class in `wizard-step-header.component.less`?
If you sure, that we do not need this class anymore.

PR #372, opened 2021-07-01
NUI-4708 docs for two directional stacked area chart

Blakesalot:


packages/charts/.gitignore
```diff
@@ -3,3 +3,4 @@ sdk/
 dist/
 **/coverage/
 .tmp
+_snapshots
```
madelyneh, line 6: Just curious, but what are they snapshots?

Blakesalot: This is from before the merge from main which already has this change. Lukas added this in his PR. It's the directory that gets created when the Percy visual test build task is run.

packages/charts/examples/components/demo/chart-types/area/chart-docs-area.module.ts
```diff
@@ -18,6 +20,7 @@ import { AreaChartVariableBaselineExampleComponent } from "./area-chart-variable
 import { AreaChartVerticalExampleComponent } from "./area-chart-vertical/area-chart-vertical-example.component";
 import { ChartDocsAreaComponent } from "./chart-docs-area.component";
 import { AreaChartTestComponent } from "./test/area-chart-test.component";
+/* eslint-enable max-len */
```
madelyneh, line 23: Isn't this on globally? why is it needed here?

madelyneh (approved):

baggiosw (approved):

packages/charts/examples/components/demo/chart-types/area/chart-docs-area.module.ts
Blakesalot: I actually disable it at the top of the file. This is just re-enabling it afterwards.

madelyneh (approved):

lukasholcik (approved):

PR #375, opened 2021-07-01
bugfix/NUI-6034 nui table in firefox column sizes

Blakesalot:


packages/bits/src/lib/table/table-state-handler.service.ts
```diff
@@ -340,7 +339,7 @@ export class TableStateHandlerService {
                 return this.state.columnsTypes[columnIndex] === "icon";
             })
             .forEach(column => {
-                this.state.columnsWidths[column] = { width: ICON_CELL_WIDTH_PX };
+                this.state.columnsWidths[column] = { width: ICON_CELL_WIDTH_PX, fixed: true };
```
Blakesalot, line 342: This makes no difference, but it's more correct.

madelyneh (approved):

lukasholcik (approved):

Blakesalot:
Closing this and rolling back the original change. We'll explore a fix later.

PR #385, opened 2021-07-08
NUI-6436 Soft Angular v12 upgrade

lukasholcik:


packages/bits/src/lib/table/table.component.ts
```diff
@@ -93,8 +110,23 @@ export class TableComponent<T> extends CdkTable<T> implements OnInit, AfterViewI
...
                 @Inject(DOCUMENT) private document: Document,
-                private platform: Platform) {
-        super(_differs, _changeDetectorRef, _elementRef, role, _dir, document, platform);
+                private platform: Platform,
+                @Inject(_VIEW_REPEATER_STRATEGY) viewRepeater: _ViewRepeater<T, RenderRow<T>, RowContext<T>>,
+                @Inject(_COALESCED_STYLE_SCHEDULER) coalescedStyleScheduler: _CoalescedStyleScheduler,
+                viewportRuler: ViewportRuler,
+                @Optional() @SkipSelf() @Inject(STICKY_POSITIONING_LISTENER) stickyPositioningListener: StickyPositioningListener
+    ) {
+        // the parameters that were added to the constructor were optional before v12
```
Blakesalot, line 119: Maybe this?
```suggestion
        // The _ViewRepeater and _CoalescedStyleScheduler parameters were optional before v12. They're included here for compatibility with v12 and later.
```

packages/bits/src/lib/table/table.component.ts
```diff
@@ -93,8 +110,23 @@ export class TableComponent<T> extends CdkTable<T> implements OnInit, AfterViewI
...
+        super(
+            _differs, _changeDetectorRef, _elementRef, role, _dir, document, platform,
+            viewRepeater, coalescedStyleScheduler, undefined, undefined
+        );
+
+        // v12 is changing the order of the last two parameters compared to v11. They deprecated the optionality of the last parameter, but the one before
```
Blakesalot, line 125: Maybe just for clarity:
```suggestion
        // Angular v12 is changing the order of the last two parameters compared to v11. They deprecated the optionality of the last parameter, but the one before
```

Blakesalot (dismissed):

packages/bits/src/lib/table/table.component.ts
```diff
@@ -93,8 +110,23 @@ export class TableComponent<T> extends CdkTable<T> implements OnInit, AfterViewI
...
+                viewportRuler: ViewportRuler,
+                @Optional() @SkipSelf() @Inject(STICKY_POSITIONING_LISTENER) stickyPositioningListener: StickyPositioningListener
+    ) {
+        // The _ViewRepeater and _CoalescedStyleScheduler parameters were optional before Angular v12. They're included here for compatibility with Angular v12 and later.
+        super(
+            _differs, _changeDetectorRef, _elementRef, role, _dir, document, platform,
+            viewRepeater, coalescedStyleScheduler, undefined, undefined
+        );
+
+        // Angular v12 is changing the order of the last two parameters compared to v11. They deprecated the optionality of the last parameter, but the one before
```
justindoherty, line 125: This is disappointing and I guess you guys can tell me I told you so. Thanks for working around it. It's a real shame and puzzling why they are doing this but also I guess I don't really know how to solve a constructor change like this in a non-breaking way.

lukasholcik: I think it all comes down to what they consider an API that feel obliged not to change. I bet there is a bunch of exposed classes that they change anyway and only consider component inputs/outputs and the really exposed things when adhering to their deprecation policy.

It might also be that CDK is not playing by the same rules as Angular core. CDK table broke during the last upgrade as well, so it's probably going to be the case regularly with this component.

That said I feel really lucky that we were able to hack our way around this. Being able to run a constructor with parameters in a different order is a tricky thing to do. Luckily enough they're just assigning these parameters to internal properties.

Blakesalot (approved):

baggiosw (approved):

justindoherty (approved):

PR #398, opened 2021-08-23
CORE-18359 | (fix) WizardStepComponent/stepTemplate

mzknksh:


packages/bits/src/lib/wizard/wizard-step.component.html
```diff
@@ -5,8 +5,8 @@
         <h2 class="nui-wizard__step-title">{{title}}</h2>
         <ng-container *ngTemplateOutlet="stepTemplate"></ng-container>
     </div>
-    <nui-spinner [percent]="busyConfig.percent"
-                 [allowCancel]="busyConfig.allowCancel"
-                 [message]="busyConfig.message">
+    <nui-spinner [percent]="busyConfig.percent || 0"
```
Blakesalot, line 8: It looks like spinner actually uses an undefined value for the `percent` input to know if it should be in "indeterminate" mode. So, it might be good to leave this as it was if it's not causing problems.

https://github.com/solarwinds/nova/blob/0a40bb0b469eb86c98ff824523b2d6201f1988af/packages/bits/src/lib/spinner/spinner.component.ts#L113

packages/bits/src/lib/wizard/wizard-step.component.html
```diff
@@ -5,8 +5,8 @@
...
+                 [allowCancel]="busyConfig.allowCancel || false"
+                 [message]="busyConfig.message || ''">
```
Blakesalot, line 10: It looks like spinner uses an undefined value for the `message` input in an `*ngIf` check. So, this is another place where it might be good to leave this as it was if it's not causing problems.

https://github.com/solarwinds/nova/blob/0a40bb0b469eb86c98ff824523b2d6201f1988af/packages/bits/src/lib/spinner/spinner.component.html#L29

mzknksh: it's just weird that the types of these input properties do not reflect that - I should probably change those accordingly so it is obvious for consumers it can be undefined

baggiosw (approved):

PR #412, opened 2021-12-01
Upgrade/ng13

mzknksh:


tomascubeek:
I hope you know that we should support one version back and we should give a notice that we plan to do such breaking change. It means that Nova should support v12 and v13 ... it also means that Nova must be compiled against v12, not v13, because if you use v13, it won't be compatible with ng v12 consumers...

More about strategy: 
https://cp.solarwinds.com/display/~tomas.kubik/Nova+Framework+Forward+Compatibility+Impact

packages/bits/angular.json
```diff
@@ -286,7 +308,6 @@
   "schematics": {
     "@schematics/angular:component": {
       "prefix": "nui",
-      "styleext": "less",
```
pavlo-poimanov, line 289: Why you removed less?

stepanstava: This property is no longer allowed.

stepanstava (dismissed):

findo11 (dismissed):

findo11 (approved):

PR #425, opened 2022-02-24
NUI-6163 | (feature) preventRowClick for nui-repeat

mzknksh:
- add implementation
- fix some repeater typings
- improve examples

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -537,9 +537,9 @@ jobs:
         environment:
             SOURCE_BRANCH: main
             # TODO: Change to false
-            CREATE_RELEASE_BRANCH: true
+            CREATE_RELEASE_BRANCH: false
             RELEASE_BRANCH: release/v11.x
-            INCREMENT_TYPE: minor
+            INCREMENT_TYPE: patch
         steps:
             -   checkout:
                     path: ~/nova
@@ -682,7 +682,7 @@ workflows:
                                 - main
                                 - /release\/.*/
                                 - next
-                                
+
     nova-release:
         jobs:
             -   release-approval:
--- a/packages/bits/src/lib/repeat/repeat.component.less
+++ b/packages/bits/src/lib/repeat/repeat.component.less
@@ -8,6 +8,10 @@
         padding-bottom: @nui-space-xs;
     }
 
+    > ul > cdk-virtual-scroll-viewport {
+        height: 100%;
+    }
+
     &__compact .nui-repeat-item {
         padding-top: 0;
         padding-bottom: 0;
@@ -18,8 +22,10 @@
         .setCssVariable(background-color, nui-color-bg-light);
     }
 
-    .repeat-group .nui-repeatitem {
+    .repeat-group .nui-repeat-item {
+        cursor: auto;
         &.nui-repeat-item--clickable {
+            cursor: pointer;
             &:hover:not(.nui-repeat-item--selected) {
                 .setCssVariable(border-color, nui-color-line-default);
             }
@@ -64,7 +70,7 @@
             }
         }
 
-        &__disabled {
+        &--disabled {
             .setCssVariable(color, nui-color-text-disabled);
             &:hover  {
                 cursor: @cursor-disabled;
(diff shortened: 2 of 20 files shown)


packages/bits/src/lib/repeat/repeat.component.ts
```diff
@@ -434,9 +527,9 @@ export class RepeatComponent<T extends IRepeatItem = any> implements OnInit, OnD
         }
     }
 
-    private itemDropped(event: IDndItemDropped) {
+    private itemDropped(event: IDndItemDropped<T>) {
         // CDK retrieves incorrectly event.previousIndex so we need to compute it ourselves
-        const item = event.item.data;
```
stepanstava, line 439: is this intentional?

.circleci/config.yml
```diff
@@ -537,9 +537,9 @@ jobs:
         environment:
             SOURCE_BRANCH: main
             # TODO: Change to false
-            CREATE_RELEASE_BRANCH: true
+            CREATE_RELEASE_BRANCH: false
```
stepanstava, line 540: do we need to change this? It shouldn't create a release branch if already exists.

stepanstava (approved):

packages/bits/src/lib/repeat/repeat.component.less
```diff
@@ -64,7 +70,7 @@
             }
         }
 
-        &__disabled {
+        &--disabled {
```
ZycaR, line 73: lol, shouldn't we let the old one too? &__disabled, &--disabled { ...

packages/bits/src/lib/repeat/repeat.component.ts
```diff
@@ -313,7 +360,7 @@ export class RepeatComponent<T extends IRepeatItem = any> implements OnInit, OnD
             type: nameof<INovaFilters>("selection"),
             value: {
                 selection: this.selection,
-                itemsSource: this.itemsSource,
+                itemsSource: this.itemsSource.slice(),
```
ZycaR, line 363: is this change needed?

packages/bits/src/lib/repeat/repeat.component.ts
```diff
@@ -349,49 +400,81 @@ export class RepeatComponent<T extends IRepeatItem = any> implements OnInit, OnD
...
+            );
         }
         this.selectionChange.emit(this.selection);
     }
 
+    public isMultiselectable(): boolean {
+        return this.selectionMode === RepeatSelectionMode.multi;
+    }
+
     /* START - ITEM BEHAVIOUR DECIDERS */
```
ZycaR, line 465: this comment should be on ln. 460

packages/bits/src/lib/repeat/repeat.component.ts
```diff
@@ -349,49 +400,81 @@ export class RepeatComponent<T extends IRepeatItem = any> implements OnInit, OnD
...
+    public isItemClickable(item: T): boolean {
+        return !this.preventRowClick && !this.isItemDisabled(item);
+    }
+
+    public isItemSelectable(item: T): boolean {
```
ZycaR, line 470: do we need this BEHAVIOUR DECIDER? 
if yes then we can use it in ln 467 too :P

ZycaR (changes requested):

packages/bits/src/lib/repeat/repeat.component.less
mzknksh: I just wanted to unify the names (other classes use --), it's internal for the component so shouldn't be an issue

packages/bits/src/lib/repeat/repeat.component.ts
mzknksh: actually leftover from change I then removed as "leave for v12" (changing of input signature to readonly array so we do not require modifiable array for itemsSource, b/c we have no reason to manipulate the array given to us, ideally all the types should change but that would be getting too ahead). I can remove this as well and leave for the v12 change

packages/bits/src/lib/repeat/repeat.component.ts
mzknksh: isMultiselectable is not item behavior but whole list behavior, item behavior deciders receive item as parameter

packages/bits/src/lib/repeat/repeat.component.ts
mzknksh: I didn't want to tie it together too much, yet have some distinct name, but I can remove it

stepanstava (dismissed):

ZycaR (approved):

stepanstava (approved):

PR #432, opened 2022-03-18
(bugfix) tooltip doesnt appear

AlekseiZhabitenko:


mzknksh (approved):

packages/bits/src/lib/tooltip/tooltip.component.ts
```diff
@@ -199,7 +199,7 @@ export class TooltipComponent implements OnDestroy, OnInit {
             .withScrollableContainers(scrollableAncestors)
             .withPositions(this.possiblePositions);
 
-        strategy.positionChanges.pipe(takeUntil(this.destroy$)).subscribe(change => {
+        strategy.positionChanges.pipe(skip(1), takeUntil(this.destroy$)).subscribe(change => {
```
mzknksh, line 202: in the meantime please also add test to verify the fix worked (and to prevent future regression)

mzknksh, line 202: in the meantime please also add test to verify the fix worked (and to prevent future regression)

AlekseiZhabitenko: Added test

mzknksh (approved):

PR #443, opened 2022-03-30
NUI-6169 Remove lodash template

erzik987:
Would be nice to create example of KPI widget with NOVA_URL_INTERACTION_HANDLER.
Ticket for that: https://jira.solarwinds.com/browse/NUI-6172


findo11 (dismissed):

packages/dashboards/src/lib/services/url-interaction.service.ts
```diff
@@ -0,0 +1,38 @@
+import { Injectable } from "@angular/core";
+
+@Injectable({
+    providedIn: "root",
+})
+export class UrlInteractionService {
+    public template(url: string, data: any): string{
+        const regex = new RegExp(/(\$\{[a-zA-Z0-9.]*\})/g)
```
mzknksh, line 8: you could have actually used capturing group for the content not for the whole expression, that would save you the slice (and it would be driven by the regex so if this ever changed you wouldn't need to manipulate the value "manually", just use the value from first captured group

packages/dashboards/src/lib/services/url-interaction.service.ts
```diff
@@ -0,0 +1,38 @@
...
+            propertyArray.push(element.slice(2,-1))
+        });
+
+        let evaluatedUrl = url;
+        for (let i = 0; i < propertyArray.length; i++) {
+            const evaluation = this.evaluate(propertyArray[i].split("."), data);
+            evaluatedUrl = evaluatedUrl.replace(interpolations[i], evaluation)
+        }
+
+        return evaluatedUrl;
```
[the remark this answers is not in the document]
erzik987: Very useful advice, thanks :)

packages/dashboards/src/lib/services/url-interaction.service.ts
erzik987: Not really sure how to do it, also I am capturing it also with parenthesis so I could replace whole expression.

packages/dashboards/src/lib/services/url-interaction.service.ts
```diff
@@ -0,0 +1,21 @@
...
+    providedIn: "root",
+})
+export class UrlInteractionService {
+    public template(url: string, data: any): string{
+        const regex = new RegExp(/\$\{([a-zA-Z0-9.]*)\}/g)
+        return url.replace(regex, (match, captured) => this.evaluate(captured.split("."), data));
+    }
+
+    private evaluate(expresion: string[], data: any): string {
+        let result = data;
```
mzknksh, line 13: we could even use `return expression.reduce((result, item) => result[item], data)` but I guess it might be less readable for someone

mzknksh (approved):

packages/dashboards/src/lib/services/url-interaction.service.ts
erzik987: Implemented :)

mzknksh (approved):

findo11 (approved):

PR #449, opened 2022-05-02
NUI-6169 Fix url-interaction-handler implementation

erzik987:


packages/dashboards/src/lib/components/providers/interaction/url-interaction-handler.ts
```diff
@@ -17,11 +17,12 @@ export class WindowObject extends Window {
...
 @Injectable()
 export class UrlInteractionHandler extends InteractionHandler<IUrlInteractionHandlerProperties, any> {
+    private urlInteractionService: UrlInteractionService
     constructor(@Inject(PIZZAGNA_EVENT_BUS) eventBus: EventBus<IEvent>,
                 @Inject("windowObject") private window: WindowObject,
-                private logger: LoggerService,
-                private urlInteractionService: UrlInteractionService) {
+                private logger: LoggerService) {
         super(eventBus);
+        this.urlInteractionService = new UrlInteractionService();
```
mzknksh, line 25: not particularly happy about this pattern - if we want it simple, might turn it into a utility function. technically the class does not have any dependencies so it's not a big deal, but I believe the core issue is elsewhere and we're just sidestepping it with avoiding DI.

erzik987: You were right, it looks like I forget to register this service into provider-registry

mzknksh (approved):

stepanstava (approved):

PR #454, opened 2022-06-28
Feature/NUI-6180-add-default-message-to-toolbar-when-no-items-present

mzknksh:
- add possibility to include message in toolbar
- add default message for toolbar without any items


packages/bits/src/lib/toolbar/toolbar.component.html
```diff
@@ -1,44 +1,92 @@
...
+                <button
+                    nui-button
+                    type="button"
+                    #toolbarButtons
+                    [tabindex]="!i && !j ? 0 : -1"
+                    (click)="onClickToolbarBtn($event, commandItem)"
+                    [displayStyle]="
+                        commandItem.displayStyle === 'main'
+                            ? 'primary'
+                            : commandItem.displayStyle
```
stepanstava, line 31: tbh I am not a big fan of this formatting in the templates, I think it's much clearer if it stays one attribute per line, even if the line is a slightly longer.

stepanstava (approved):

PR #467, opened 2022-08-02
CORE-18288 Add default value for conversion pipe

erzik987:


mzknksh (dismissed):

ales-fuks (dismissed):

packages/dashboards/src/lib/common/pipes/dashboard-unit-conversion-pipe.ts
```diff
@@ -17,10 +17,14 @@ export class DashboardUnitConversionPipe implements PipeTransform {
      *
      * @returns The string representation of the converted value
      */
-    public transform = (value: string | number | undefined, units: UnitOption = "generic"): string => {
+    public transform = (
+        value: string | number | undefined,
+        units: UnitOption = "generic",
+        defaultThreshhold: number = DEFAULT_UNIT_CONVERSION_THRESHOLD
```
ZycaR, line 23: you have typo in name of the argument (defaultThreshold) and you should use the enum we already have :UnitBase:

similar to this:
https://github.com/solarwinds/nova/blob/c82170c479d4579c0a2f4e0e94ce8b705823f973/packages/bits/src/services/unit-conversion.service.ts#L26

packages/dashboards/src/lib/components/timeseries-widget/timeseries-scales.service.ts
```diff
@@ -32,7 +32,7 @@ export class TimeseriesScalesService {
             }
             case TimeseriesScaleType.Linear: {
                 scale = new LinearScale();
-                scale.formatters.tick = (value: string | number | undefined) => this.unitConversionPipe.transform(value, units);
+                scale.formatters.tick = (value: string | number | undefined) => this.unitConversionPipe.transform(value, units, 1000);
```
ZycaR, line 35: UnitBase.Standard instead of 1000.

ZycaR (changes requested):

ales-fuks (approved):

findo11 (approved):

PR #490, opened 2022-11-08
alllow timeserieswidget to specify collection id

mzknksh:
+ hack x scales on interaction plugins
- still missing proper marks on non identical x-scales


packages/dashboards/package.json
```diff
@@ -148,6 +146,12 @@
         "serve-examples": "ng s examples",
         "serve-examples:prod": "npx local-web-server -p 4200 -d sdk/api-docs-ng2/examples --spa index.html",
         "start": "run-p -lr compodoc:serve build-examples:watch",
+        "symlink": "run-s symlink:*",
```
stepanstava, line 149: I guess you need to add it to "assemble" or are you going to change the build config?

packages/dashboards/package.json
```diff
@@ -148,6 +146,12 @@
...
+        "symlink:bits": "npx rimraf node_modules/@nova-ui/bits && npx symlink-dir ../bits/dist/ node_modules/@nova-ui/bits",
+        "symlink:charts": "npx rimraf node_modules/@nova-ui/charts && npx symlink-dir ../charts/dist/ node_modules/@nova-ui/charts",
+        "symunlink": "run-s symlink_undo:*",
```
stepanstava, line 152: no symlink_undo command

packages/charts/package.json
```diff
@@ -103,6 +102,7 @@
     },
     "scripts": {
         "assemble": "run-s prettier-check lint build-lib:locl build-lib:prod compodoc-custom-theme build-examples:prod",
+        "assemble-dev": "yarn run build-lib:prod",
```
stepanstava, line 105: for "assemble-dev" would be better to run build-lib with non prod config

stepanstava (commented):
lgtm

packages/charts/package.json
mzknksh: the intention is to have the same output as regular assemble without running too many things repeatedly (to be used when working on some feature and testing it's use in some linked project)

mzknksh: but as long as dist-copy is part of build-lib:prod I can just remove it and use that one, I wanted to keep the dist-copy only as part of assemble scripts but I guess I will leave it as it was in the end

stepanstava (approved):

PR #493, opened 2022-11-16
fix chart collections

mzknksh:


packages/charts/src/chart-collection/chart-collection-id.directive.ts
```diff
@@ -42,9 +48,21 @@ export class ChartCollectionIdDirective implements AfterViewInit {
...
 
-    public ngAfterViewInit(): void {
-        this.chartCollectionService
-            .getChartCollection(this.collectionId)
-            .addChart(this.chartComponent.chart);
+    public ngOnChanges(changes: SimpleChanges): void {
+        console.log("ngOnChanges");
+        const collectionIdChange = changes.collectionId;
+        if (collectionIdChange) {
+            console.log("collectionIdChange", collectionIdChange);
```
stepanstava, line 55: please remove console logs

PR #518, opened 2023-01-25
oo-8567 timeseries legend menu

findo11:


packages/dashboards/src/lib/components/timeseries-widget/chart-presets/timeseries-chart.component.ts
```diff
@@ -152,6 +163,19 @@ export abstract class TimeseriesChartComponent<T = ITimeseriesWidgetSeriesData>
...
+                // save original data and transform it
+                this.widgetData.series.forEach((serie) => {
+                    serie.rawData = serie.data;
+                    serie.transformer =
+                        changes.widgetData.previousValue?.series.find(
+                            (prevSerie: ITimeseriesWidgetData) =>
+                                prevSerie.id === serie.id
+                        )?.transformer;
+                    if (serie.transformer && serie.rawData.length > 0) {
+                        serie.data = serie.transformer(serie.rawData);
```
stepanstava, line 175: same here, could we do the copy only in the case that the transormation will be applied?

````
if (serie.transformer && serie.data.length > 0) {
  serie.rawData = [...serie.data];
  serie.data = serie.transformer(serie.data);
}
````

**Review by stepanstava — CHANGES_REQUESTED** [review]

**Review by stepanstava — DISMISSED** [review]

**Review by stepanstava — APPROVED** [review]

File: packages/dashboards/src/lib/components/timeseries-widget/transformer/loess.ts [hunk 2]
```diff
@@ -0,0 +1,283 @@
+// Based on org.apache.commons.math.analysis.interpolation.LoessInterpolator
```
mzknksh, line 1: we should also include the comment header to newly added the source files (keep these, but prepend the SWI text as seen in other sourcefiles)

(did we check the original license in these? just to be sure)

mzknksh (changes requested):

mzknksh (approved):

stepanstava (approved):

PR #529, opened 2023-02-24
NUI-6196 | add declarative radio hints

mzknksh:


stepanstava:
do we need to introduce a new nui-radio-hint component whose purpose is only to project the content?

I think it would be better to extend the functionality of nui-radio by adding a new hintTemplate input that would do the same.

Similarly as FormFieldComponent has hint and hintTemplate:
https://nova-ui.solarwinds.io/bits/release_v11.x/components/FormFieldComponent.html (Hint Text section)

mzknksh:
originally I did it with a selector only, then I wanted to try the click propagation prevention for links, but that could be a general reusable directive instead

not sure if having to supply extra template for each item is a better option, but I guess having some kind of consistency would be useful

pavlo-poimanov (dismissed):

PR #540, opened 2023-03-17
Forces rounding of numbers with many decimals.

michal-aron-swi:


stepanstava (approved):

packages/dashboards/src/lib/components/timeseries-widget/timeseries-scales.service.ts
```diff
@@ -224,6 +224,32 @@ export class TimeseriesScalesService {
...
+        return this.roundToOptimalDecimals([
+            min,
+            min + point,
+            min + 2 * point,
+            max - point,
+            max,
+        ]);
+    }
+
+    private roundToOptimalDecimals(arr: number[]): number[] {
```
mzknksh, line 236: what about something like this?

```
const optimalDecimals = new Map<number, number>([
    [1, 2],
    [2, 2],
    [3, 2],
    [4, 1],
]);

const roundToOptimalDecimals = (numbers: number[]) => numbers.map(n => {
        const digits = Math.floor(Math.log10(Math.abs(n))) + 1;
        const optimalDecimalsForDigits = optimalDecimals.get(digits) ?? 0;
        return n.toFixed(optimalDecimalsForDigits);
    });
```

PR #570, opened 2023-05-23
OO-19330 chart units v2

KleinMichalGit:


pavlo-poimanov:
please use yarn to update lock

PR #623, opened 2023-07-20
update d3 library from v5 to v7

KleinMichalGit:


KleinMichalGit:
FIX: all d3 lib versions were changed according to d3 lib versions in Fusion. Also, yarn.lock in charts was deleted and reinstalled

findo11 (approved):

PR #627, opened 2023-08-10
Fusion Migration - Perfstack - Missing remove button in summary series

pilat-martin:


ludmilaFialova (approved):

packages/dashboards/src/lib/components/timeseries-widget/chart-presets/xy-chart/xy-chart.component.html
```diff
@@ -213,15 +213,6 @@
                         </nui-menu-action>
                     </nui-menu-group>
                 </nui-menu>
-                <button
-                    nui-button
-                    legendMenu
-                    icon="remove"
-                    type="button"
-                    *ngIf="displayDeleteButton()"
```
pavlo-poimanov, line 221: why this one is deleted but not used the ngIf condition? 
shoudn't the condition be removed as well?

pilat-martin: I removed the button because it is not supposed to be there:
image

The displayDeleteButton() is kept because it is used on line 77 as well (if thats what you were referring to).

Issue #639, opened 2023-09-14
Add RiskScore Component to the Bits package


MykhailoMolyboha:
## Related Jira item

https://swicloud.atlassian.net/browse/OO-23867

## Description
For SecObs we need to add a RiskScore Component and after that RiskScore widget.

## Screenshots (if applicable)
mockup
![image](https://github.com/solarwinds/orion/assets/40260984/89c8d8b7-04ca-4c86-a38b-82399119045b)

PR #648, opened 2023-09-15
Add RiskScore Component to the Bits package #639

MykhailoMolyboha:
## Related Jira item

https://swicloud.atlassian.net/browse/OO-23867

## Issue
https://github.com/solarwinds/nova/issues/639

## Description
Add Fusion shared 'Risk Score' component

## Screenshots (if applicable)
mockup
image

/components/RiskScoreComponent.html#example
image

## Definition of done

- [x] Unit and E2E tests are written (or will be in scope of another PR)
- [x] Feature toggling is applied - Not Applicable
- [ ] Performance impact was considered
- [x] The security and privacy aspect has been considered - Not Applicable
- [x] Demo mode is considered - Not Applicable

### DIFF
--- a/packages/bits/demo/src/components/app/app-routing.module.ts
+++ b/packages/bits/demo/src/components/app/app-routing.module.ts
@@ -203,6 +203,10 @@ const appRoutes: Routes = [
             import("../demo/runtime-i18n/runtime-i18n.module"),
     },
     {
+        path: "risk-score",
+        loadChildren: async () => import("../demo/risk-score/risk-score.module"),
+    },
+    {
         path: "search",
         loadChildren: async () => import("../demo/search/search.module"),
     },
--- a/packages/bits/demo/src/components/app/app-routing.module.ts
+++ b/packages/bits/demo/src/components/app/app-routing.module.ts
@@ -204,7 +204,8 @@ const appRoutes: Routes = [
     },
     {
         path: "risk-score",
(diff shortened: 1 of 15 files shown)


Blakesalot:
@MykhailoMolyboha, I'm no longer a Nova maintainer so I have no official say in this, but in my opinion this component is very feature-specific and should live somewhere else besides the Nova libraries. The set of components in Nova has been carefully curated to ensure that they all meet a minimum amount of general usefulness before being included. In this case, I would recommend adding it to the specific project that needs it and reuse it from there, but I'll defer to the existing package owners on the final decision.

Blakesalot:
@MykhailoMolyboha, I should point out that Nova Charts does have a gauge component with radial (donut) and linear modes that was intended for use cases similar to this.

Here are the docs for that: https://nova-ui.solarwinds.io/#/charts/main/additional-documentation/chart-types/gauge.html
And, here is the source code: https://github.com/solarwinds/nova/tree/main/packages/charts/src/gauge

The look and feel are tailored to the Nova design spec, so it doesn’t **look** the same as the component in this PR, but it offers capabilities similar to those needed by your use case. And, it could potentially be enhanced to fulfill any requirements that it currently doesn’t meet.

findo11 (approved):

findo11:
Approved, but please consider Blake's idea.

PR #866, opened 2026-06-04
Bugfix/oo 50848/nui toolbar access new

harsha573252:
## Frontend Pull Request Description
##Related Jira
https://swicloud.atlassian.net/browse/OO-50848
Fixes:
(a11y): add ARIA labels, Home/End navigation, decorative button icons and updated version

- Added aria-label/aria-labelledby
- Added Home & End keyboard support with tests
- Set button icons as decorative ([decorative]="true")
- Updated version to 20.0.6 across all packages

## Checklist

- [ ] My code follows the [style guidelines](https://github.com/solarwinds/nova/blob/main/docs/STYLE_GUIDE.md) of this project
- [ ] I have performed a self-review of my code
- [ ] I have updated [change log](https://github.com/solarwinds/nova/blob/main/docs/CHANGELOG.md)
- [ ] I have been following [Definition of done](https://github.com/solarwinds/nova/blob/main/docs/DEFINITION_OF_DONE.md)
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new lint warnings
- [ ] New and existing unit tests pass locally and on CI with my changes
- [ ] Any dependent changes have been merged and published in downstream modules


ludmilaFialova (dismissed):

packages/bits/src/lib/toolbar/toolbar-keyboard.service.ts
ludmilaFialova: if there is a possibility to have a vertical toolbar, we need also up and down arrows...

packages/bits/src/lib/toolbar/toolbar.component.ts
```diff
@@ -78,6 +81,12 @@ export class ToolbarComponent implements AfterViewInit, OnDestroy {
     @ContentChildren(ToolbarItemComponent, { descendants: true })
     public items: QueryList<ToolbarItemComponent>;
```
ludmilaFialova, line 82: I don't see any logic for skipping disabled items.

packages/bits/src/lib/toolbar/toolbar-keyboard.service.spec.ts
```diff
@@ -122,6 +122,36 @@ describe("Services > ", () => {
...
+                const spy = spyOn(last, "focus");
+
+                first.focus();
+
+                service.onKeyDown({
+                    ...keyboardEventMock,
+                    code: KEYBOARD_CODE.END,
+                } as KeyboardEvent);
+                expect(spy).toHaveBeenCalled();
+            });
```
ludmilaFialova, line 154: I would add some tests if aria-label etc. are not missing in DOM.

packages/bits/src/lib/toolbar/toolbar-keyboard.service.ts
harsha573252: Added Up/Down arrow support for vertical toolbars

packages/bits/src/lib/toolbar/toolbar.component.ts
harsha573252: Added disabled-item skipping in keyboard navigation.

packages/bits/src/lib/toolbar/toolbar-keyboard.service.spec.ts
harsha573252: Added aria-label/orientation DOM tests.

packages/bits/src/lib/toolbar/toolbar.component.html
```diff
@@ -24,6 +24,7 @@
                     type="button"
                     #toolbarButtons
                     [tabindex]="!i && !j ? 0 : -1"
```
ludmilaFialova: The first item always receives tabindex="0", even when it is disabled. Since disabled buttons cannot receive focus and all remaining items have tabindex="-1", the entire toolbar may become unreachable via Tab.

Please assign tabindex="0" to the first enabled item instead. If no enabled toolbar items exist, the overflow menu toggle should be the tab stop. Please also add a test covering a disabled first item.

packages/bits/src/lib/toolbar/toolbar.component.ts
ludmilaFialova: again the issue with disabled buttons

packages/bits/src/lib/toolbar/toolbar.component.html
ludmilaFialova: A disabled toolbar item becomes actionable when moved to the overflow menu because its disabled state is not passed to nui-menu-action.

Please add [disabled]="menuItem.disabled" and handle activation through (actionDone) instead of (click), so disabled items cannot emit the action.

Please also add a test covering a disabled item in the overflow menu.

packages/bits/src/lib/toolbar/toolbar.component.ts
```diff
@@ -326,7 +335,8 @@ export class ToolbarComponent implements AfterViewInit, OnDestroy {
 
                 this.keyboardService.setToolbarItems(
                     this.toolbarItems,
-                    this.menu
+                    this.menu,
+                    this.ariaOrientation
```
ludmilaFialova, line 367: If ariaOrientation changes at runtime, the ARIA attribute updates, but the keyboard service may retain the previous orientation because it is only synchronized when the button list changes. This can cause the announced orientation and arrow-key behavior to differ.

Please synchronize the service whenever ariaOrientation changes, for example via an input setter or ngOnChanges, and add a test for a runtime orientation change.

ludmilaFialova (dismissed):

ludmilaFialova (dismissed):

ludmilaFialova (approved):

PR #877, opened 2026-06-22
Added unique IDs to expander landmarks

harsha573252:
# Releated Jira
https://swicloud.atlassian.net/browse/OO-50773

## Frontend Pull Request Description
Expander: Add unique IDs to region landmarks with aria-label

## Checklist

- [ ] My code follows the [style guidelines](https://github.com/solarwinds/nova/blob/main/docs/STYLE_GUIDE.md) of this project
- [ ] I have performed a self-review of my code
- [ ] I have updated [change log](https://github.com/solarwinds/nova/blob/main/docs/CHANGELOG.md)
- [ ] I have been following [Definition of done](https://github.com/solarwinds/nova/blob/main/docs/DEFINITION_OF_DONE.md)
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new lint warnings
- [ ] New and existing unit tests pass locally and on CI with my changes
- [ ] Any dependent changes have been merged and published in downstream modules


packages/bits/src/lib/expander/expander.component.html
```diff
@@ -50,7 +51,7 @@
     <div
         class="nui-expander__body-wrapper"
         role="region"
-        [attr.aria-label]="header || null"
+        [attr.aria-labelledby]="'nui-expander-header-' + uniqueId"
```
ludmilaFialova: Using the whole header element as the aria-labelledby target can make the region inherit the header’s action-oriented label (e.g. “Archive toggle”), not just the section name.
Could we reference a dedicated label element (text/content only) instead, and point both the button and the region to that same label ID?

[attr.aria-labelledby]="'nui-expander-label-' + uniqueId"

packages/bits/src/lib/expander/expander.component.html
```diff
@@ -3,7 +3,8 @@
         class="nui-expander__header"
         tabindex="0"
         role="button"
-        aria-label="Expander toggle"
+        [id]="'nui-expander-header-' + uniqueId"
+        [attr.aria-label]="accessibleLabel + ' toggle'"
```
ludmilaFialova:
```
-    [id]="'nui-expander-header-' + uniqueId"
-    [attr.aria-label]="accessibleLabel + ' toggle'"
+    [attr.aria-labelledby]="'nui-expander-label-' + uniqueId"
```

packages/bits/src/lib/expander/expander.component.ts
```diff
@@ -84,10 +90,20 @@ export class ExpanderComponent implements AfterContentInit {
...
-    constructor(private cdRef: ChangeDetectorRef) {}
+    constructor(private cdRef: ChangeDetectorRef) {
+        this.uniqueId = String(++nextExpanderId);
+    }
+
+    /**
+     * Accessible name for the toggle button and region landmark.
+     */
+    public get accessibleLabel(): string {
+        return this.header || this.ariaLabel || `Expander ${this.uniqueId}`;
```
ludmilaFialova: With the shared aria-labelledby approach (button + region pointing to a dedicated label element), it seems we could remove this method and rely on the label element content (including ariaLabel as SR-only fallback when no visible/custom header is present). This would simplify naming logic and avoid the hardcoded Expander ${uniqueId} fallback path.

packages/bits/src/lib/expander/expander.component.html
ludmilaFialova, line 47: Here we can add a screen-reader-only fallback label when the expander has no visible header but receives ariaLabel. This keeps the accessible name available to both the toggle and the region without changing the visual layout:

            <span
                *ngIf="!header && ariaLabel"
                class="nui-sr-only"
            >
                {{ ariaLabel }}
            </span>

packages/bits/src/lib/expander/expander.component.ts
ludmilaFialova, line 72: Please add focused a11y tests for the aria-labelledby relationship and this ariaLabel fallback, including standard, custom, and headerless expanders

harsha573252: added tests :)

ludmilaFialova (dismissed):

ludmilaFialova (dismissed):

ludmilaFialova (dismissed):

ludmilaFialova (approved):

